PR #28, opened 2020-02-04
Removed Python version matrix

bfc5288:
Removed Python version matrix and used Py3.6 instead.


myNameIsPatrick:
I'm going to merge this in, since the only issue was because of a repeated upload with the same version. It looks like we can't reupload the same package with the same version twice to PyPI, even if it was removed before, unfortunately.

Issue #32, opened 2020-02-12
Add requirements.txt


mjuric:
We should add a `requirements.txt` file with all the dependencies necessary to develop the package, so a dev can do:
```
pip install -r requirements.txt
```
to get themselves started. This will largely be a union of packages mentioned in `install_requires` and `extras_require` variables in `setup.py`.

P.S.: We may need to add a separate `conda-requirements.txt`, in case there are any packages where the PyPI name is different from the `conda` package name.

adambrazier:
Also it's best to specify specific known-good version numbers, rather than >=, as in a few years ago those can bite us...On Feb 11, 2020 8:24 PM, Mario Juric <notifications@github.com> wrote:We should add a requirements.txt file with all the dependencies necessary to develop the package, so a dev can do:
pip install -r requirements.txt

to get themselves started. This will largely be a union of packages mentioned in install_requires and extras_require variables in setup.py.
P.S.: We may need to add a separate conda-requirements.txt, in case there are any packages where the PyPI name is different from the conda package name.

PR #33, opened 2020-02-12
Add requirements

AbhishekSrikanth:
### This solves issue #32
- added requirements.txt

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,10 @@
 genesis==0.0.0
 pytest==5.3.5
+Sphinx==2.4.1
+pytest-cov==2.8.1
+pytest-console-scripts==0.2.0 
+flake8-black==0.1.1 
 sphinx_rtd_theme==0.4.3
+sphinxcontrib-programoutput==0.15
 adc==0.0.1
 flake8==3.7.9
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,10 +1,1 @@
-genesis==0.0.0
-pytest==5.3.5
-Sphinx==2.4.1
-pytest-cov==2.8.1
-pytest-console-scripts==0.2.0 
-flake8-black==0.1.1 
-sphinx_rtd_theme==0.4.3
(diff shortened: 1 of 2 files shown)


mjuric:
Thanks @AbhishekSrikanth!

@myNameIsPatrick, can you take a look (I'll be traveling a lot this week & don't have time right now)?

Btw., I'm not sure what the best practice is for version pinning; there are some [convincing arguments](https://blog.miguelgrinberg.com/post/the-package-dependency-blues) to fully pin in `requirements.txt` and allow float in `install_requires`, but the best practice doesn't seem fully established (e.g.,  [astroquery](https://github.com/astropy/astroquery/blob/master/pip-requirements) defines the lower bound). I admit to not knowing enough about this to strongly argue one way or another (my workflow is conda-based). @adambrazier -- any thoughts?

myNameIsPatrick:
@mjuric, no problem.

From my understanding, the choice of pinning dependencies vs. not has to do on whether I'm developing an application or a library. For a library, I wouldn't want to pin dependencies since it would need to play nice with the other libraries in my environment, and pinning makes it difficult to have other libraries coexist without running into version conflicts.

For an application, I'd want to be able to create a reproducible environment out of a list of dependencies. That may cause some issues if we're targeting multiple package managers though. You can usually find a single set of versions that work with pip + conda, but if we were targeting say, RHEL or debian as well at some point, we're going to be running into issues. Realistically, having a fully pinned environment likely means we'll be limited to running in a virtualenv or conda env.

I don't really have any strong preferences either, but the 

> fully pin in requirements.txt and allow float in install_requires"

approach sounds reasonable to me.

requirements.txt
```diff
@@ -0,0 +1,10 @@
+genesis==0.0.0
```
myNameIsPatrick: I would remove this, the genesis client is imported via `import genesis` but the PyPI package is named `adc`. That's probably why both of these are in here, but the `genesis` package refers to an unrelated package we don't want.

Issue #10, opened 2020-01-29
Automate Deployment to PyPI and anaconda.org


mjuric:
As a developer, I want a new package version to be automatically (and reproducibly) built, tested, and uploaded to PyPi and anaconda.org any time a semver-like tag (i.e., `vX.X.X`) is pushed into the repository.

Definition of done:
- [x] Pushing a new vX.X.X tag triggers `pip` and `conda` package builds, which upload to PyPI and anaconda.org if successful.
- [ ] Documentation should also be built and published to RTD (see #9)

Tasks:
- [x] Create scimma accounts for Anaconda and PyPi
- [x] Add scimma account credentials to Github Secrets
- [x] Create Github action workflow to upload to PyPi
- [x] Create Github action workflow to upload to Anaconda
- [x] Allow conda-build to extract the version number from tags (see #11)
- [x] Document the workflow usage

PR #38, opened 2020-02-20
Automate Deployment to PyPI and anaconda.org

bfc5288:
This PR implements #10 by creating a Github workflow that builds and uploads the repo as a package to PyPI and the Anaconda SCiMMA Organization channel when a commit is pushed with a new semver tag (via, e.g., `git tag v0.0.1`). The Anaconda job implements the `conda-build` versioning functionality mentioned in #11. The upload jobs run in sequence: first PyPI, then Anaconda.

The PyPI job builds using `setuptools` and `wheel`, then uploads with `twine`.

The Anaconda job updates the package recipe with 1) the scm version from `setup.py` and 2) the tarball hash generated with `yq` from the PyPI package, before building and uploading. The Anaconda job runs in a Miniconda3 container to reduce time needed for workflow setup.


myNameIsPatrick:
@bfc5288, this all looks good to me. Nice work!

I'll go approve this and then you can merge this in whenever. Moving forward, given the number of tags associated with v0.0.3 (and alphas), I think it would be reasonable to remove all of those (and the versions uploaded to PyPI/conda to avoid confusion. Any subsequent releases would start from v0.0.4 with a corresponding update to changelog, etc. whenever we have any new content (either the gcn subscriber or the auth/encryption capability, whichever comes first). Does that plan sound reasonable to you?

bfc5288:
@myNameIsPatrick, that sounds good, we should roll out to v0.0.4 on the next content addition. I removed all of the Github tags and Anaconda packages matching v0.0.3*, though you'll have to remove the packages from PyPI as the owner. Merging PR.

Issue #37, opened 2020-02-18
Authentication and encryption support


mjuric:
As a user, I want to be able to access servers requiring SSL and SASL+PLAIN authentication

Definition of done:
- [x] Be capable of accessing topics in SSL + auth-enabled kafka in the scimma-server-container.
- [x] The client must still be capable of accessing unprotected streams

PR #39, opened 2020-02-21
Authentication and encryption support

myNameIsPatrick:
This PR adds authentication and encryption support for the client as specified by #37, by allowing configuration options to be passed into the genesis client from `scimma publish`. Two types of new options are allowed to be able to configure the client, mimicking the style of `kafkacat`:

1. Pass in individual options via `-X property=value`.
2. Pass in a configuration file via `-F <config-file>`.

Additional items this PR addresses:

* Update CHANGELOG with feature addition
* Update README, in quickstart section
* Bump minimum version of genesis-client (`adc`) to 0.0.2
* Update unit tests to pass with modified client parameters.


bfc5288:
@myNameIsPatrick, this looks good. I tested passing an SSL config with `-F` as well as individual parameters with `-X`, which produce the expected behavior. The non-authenticated behavior (with the docker `--noSecurity` option) continues to work as well; approving the PR.

Issue #48, opened 2020-02-29
Expose a python-based client API


myNameIsPatrick:
As a user, I want to be able to read and write messages stored in the PubSub system using a python client API.

Tasks:

* [x] Have analogous functionality (when applicable) to `scimma publish` via python API calls
* [x] Have analogous functionality (when applicable) to `scimma subcribe` via python API calls
* [x] Write unit tests
* [x] Write documentation

Issue #5, opened 2020-01-29
Receive GCN Circulars from the PubSub system


mjuric:
As a user, I want to be able to receive GCN Circulars stored in the PubSub system in a JSON envelope (see #4). It must be possible to pretty-print these to the screen.

Definition of done:
- Running `scimma subscribe gcn` must begin listening for new messages from the topic `gcn`. Any received message can be assumed to be a JSON-serialized GCN Circular (see #4), and shall be pretty-printed to the screen (i.e., in the usual GCN Circular format of headers, followed by the body).
- Running `scimma subscribe gcn --json` will do the same as the above, except dump the message in JSON.
- `scimma subscribe` will support the following additional options:
  - `--earliest`: start streaming from the earliest available Kafka offset
  - `--timeout N`: timeout and exit if no new message is received after `N` seconds 
- Tests and documentation will exist for all new functionality

Use [genesis-client](https://github.com/astronomy-commons/genesis-client/) for low-level Kafka connection/communication (fork and modify it as necessary, if necessary).

Tasks:
- [x] Create the `scimma` entry point (may already be done in #4)
- [x] Expose a friendly argparse cmdline interface
- [x] Write unit tests
- [x] Document the new functionality

Issue #44, opened 2020-02-26
Stream-based GCN publisher via VOEvent transport protocol


myNameIsPatrick:
VOEvent transport protocol: http://www.ivoa.net/documents/VOEventTransport/20170320/REC-VTP-2.0-20170320.html

rdt12:
This is addressed by the git repo: https://github.com/scimma/gcn2scimma

Issue #43, opened 2020-02-26
Publish GCN Notices from VOEvent-formatted files


myNameIsPatrick:
Extend a command-line client to ingest GCN Notices (VOEvents) and post them using VOEvent JSON serialization. Similarly to how `scimma publish` operates, should read in XML files.

Schema: http://www.ivoa.net/documents/Notes/VOEventJSON/20171014/VOEventJSON.pdf

As for possible libraries to leverage, there's https://github.com/timstaley/voevent-parse. VOEventLib and pygcn are both GPL-licensed.

It may make more sense to split up the client into the core client library to just interact with the Kafka broker via scimma (scimma-client) and distinct sources that would ingest packets/files/etc from a variety of methods into the scimma ecosystem. In this case, we could have a library that would just handle GCN notices and leverage pygcn for this. That single library would be GPL licensed but the rest (and the core client) would still be BSD licensed without causing issues.

rdt12:
@myNameIsPatrick I have a very simple gcn reader that outputs JSON in a stand-alone repository:

https://github.com/scimma/gcn2stdout

I'm working on another program named voevent2scimma that will read input from gcn2stdout (or any other program that outputs VOEvents in  a specified format) and publish to scimma. I plan for it to take the program to use to get VOEvent messages as a configuration parameter. It will import the scimma client library and use it to interact with the scimma server.

I have a question about the format of input messages to the scimma client. Should the client accept XML and do the parsing or should it take JSON (if so, what format)?
Currently, gcn2stdout outputs JSON, but it would be easy to make it output raw XML.

myNameIsPatrick:
@rdt12, that URL returns a 404 error for me, I may not be a member of that repo if it's private.

That plan sounds reasonable to me. I'd rather the client accept JSON with the schema defined in the URL in the first comment. I see slight differences in the schema just due to how XML can have duplicate keys, but it should almost be one-to-one.

I have started work on the schemas for VOEvents and Circulars, as a way to more finely control the data products we support. The VOEvent data class is here: https://github.com/scimma/client_library/blob/master/scimma/client/models.py#L11. As long as the JSON packet is read in via `json.loads` or similar, you should be able to unpack it into the data structure by doing:

```
from scimma.client import models

msg = json.loads(msg_json)
voevent = models.VOEvent(**msg)
```

and then to send it to the client, you can serialize it like:

```
packet = json.dumps(voevent.todict())
```

[...]

PR #49, opened 2020-03-01
Expose a python-based client API

myNameIsPatrick:
## Description

This PR implements #48, to define a python-based client API to read and write messages from a Kafka broker. The way it's done is to have a small wrapper around the genesis client, allowing us to implement different defaults and further down the line, control which messages are allowed without expecting the genesis client to do so.

In doing so, I have also implemented data classes for GCN Circulars and VOEvents to define the structure of messages, which are used within `scimma publish` for the circulars.

Additionally, there has been some restructuring in the test suite to avoid repeating code.

This can merged independently of #5, and the test suite restructure may help out there as well.

One example:
```
    from scimma.client import stream
    with stream.open("kafka://hostname:port/topic", "r", format="json") as s:
        for idx, msg in s:
             print(msg)

    with stream.open("kafka://hostname:port/topic", "w", format="json") as s:
        s.write({"my": "message"})
```

## Checklist

### DIFF
--- a/doc/api/api.rst
+++ b/doc/api/api.rst
@@ -6,4 +6,6 @@ scimma-client API
 .. toctree::
     :maxdepth: 2
 
+    io
     publish
+    models
--- a/doc/api/io.rst
+++ b/doc/api/io.rst
@@ -1,0 +1,7 @@
+.. _io:
+
+scimma.client.io
+################
+
+.. automodule:: scimma.client.io
+    :members:
--- a/doc/api/models.rst
+++ b/doc/api/models.rst
@@ -1,0 +1,7 @@
+.. _models:
+
+scimma.client.models
+#####################
+
+.. automodule:: scimma.client.models
+    :members:
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -10,6 +10,7 @@ User's Guide
    user/installation
    user/quickstart
    user/commands
+   user/stream
 
 API Reference
 -------------
--- a/doc/user/quickstart.rst
+++ b/doc/user/quickstart.rst
@@ -8,6 +8,8 @@ Quickstart
 Reading messages
 ----------------
 
+The scimma client supports a python-based API for reading messages from a stream, as follows:
+
 .. code:: python
 
     from scimma.client import stream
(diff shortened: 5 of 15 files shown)


bfc5288:
This looks good from the tests -- I ran them all and it passed. But following the commands from the quickstart, I couldn't read/write messages to the broker (the Python prompt would hang). I'm not sure if this is an issue with my environment, so I'll look into this more.

I didn't test the VOEvent data model yet, which would be good to try since it isn't included in the tests (or publish.py) so far.

myNameIsPatrick:
@bfc5288, great, thanks for looking this over. As for the quickstart example, did you spin up the server with the `--noSecurity` option or with SSL enabled? There's some funny business using the client (and not limited to this, with kafkacat as well) where the client would hang if the SSL configuration was either misconfigured or due to having issues with the container.

As for the VOEvent model, I wouldn't get too hung up on that yet. I added that in there in preparation for #43 and #44, and don't have a good way of verifying the schema without making headway on those issues. I don't expect issues with the dataclass itself but possibly in how the schema is structured.

bfc5288:
I redid the quickstart with the `--noSecurity` option to remove that variable (I had it enabled during the first test). The write functionality works, as: 
```
with stream.open("kafka://scimma-server_noSec:9092/gcn", "w", format="json") as s:
    s.write({"my": "message"})
```
I can read the message with kafkcat: `kafkacat -C -b scimma-server_noSec:9092 -t topic`, but not with the read functionality: 
```
with stream.open("kafka://scimma-server_noSec:9092/topic", "r", format="json") as s:
    for idx, msg in s:
        print(msg)
```
(this hangs without returning the message as expected).

And that's fine about VOEvent -- we can look at that later when we're working with that format.

myNameIsPatrick:
Yeah, so I realize now that the quickstart would be confusing now without some context as to what's going on. There's an extra option needed if you want to read in messages sent before you spun up the consumer as follows:

```
with stream.open("kafka://localhost:9092/test", "r", format="json", start_at="earliest") as s:
    ...
```
I'll edit the quickstart accordingly and add more context so users aren't confused by the hanging feature.

PR #54, opened 2020-03-13
Publish GCN Notices from VOEvent-formatted files

myNameIsPatrick:
## Description

This PR implements #43, to be able to publish GCN notices from VOEvent files in XML format. In doing this, I had put some thought into how the command line interface was going to look like as we allow more types of messages to be published (and subscribed from) without cluttering the API and hopefully making it a bit intuitive to use.

Here's what I came up with:

```
scimma publish kafka://hostname:port/notice example1.xml example2.xml
```

I thought a reasonable thing to do was to make the parsing based on the file extension until we have proper schema validation. VOEvents (which notices are formatted in commonly) reasonably expect to be in XML format and circulars stored on the NASA GCN website have the .gcn3 extension.

I also changed the broker URL to be a positional argument, since it's really required and I think it looks a little cleaner. I also expect once we have a canonical URL we can point to with well-defined topics, someone can drop most of the URL and it'll pick up defaults as needed. Something like:

```
scimma publish notice example1.xml example2.xml
```
which when entered, should fill in the default kafka://hostname:port combination expecting to publish to the "notice" topic.

I renamed a few things and edited some comments to reflect the broader nature of publish.py.

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ SCiMMA Client
 Publish a GCN to Kafka:
 
 ```
-scimma publish -b kafka://hostname:port/gcn mygcn.gcn3
+scimma publish kafka://hostname:port/gcn mygcn.gcn3
 ```
 
 An example RFC 822 formatted GCN circular (`example.gcn3`) is provided in
--- a/doc/user/quickstart.rst
+++ b/doc/user/quickstart.rst
@@ -49,7 +49,7 @@ Publish a GCN
 
 .. code:: bash
 
-    scimma publish -b kafka://hostname:port/gcn mygcn.gcn3
+    scimma publish kafka://hostname:port/gcn mygcn.gcn3
 
 
 An example RFC 822 formatted GCN circular (:code:`example.gcn3`) is provided in :code:`tests/data`.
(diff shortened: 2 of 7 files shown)


myNameIsPatrick:
So, first error was due to a new version of pytest released today that seemed to have caused some issues with how fixtures are defined. That was fixed in the second to last commit. I also forgot to actually include the models unit tests during the last PR so I added that in.

This error in the python 3.6 job is strange, and I'll have to investigate this further. My local version of python is 3.7 so I never caught this, but it seems to be some idiosyncratic behavior in one of the unit tests.

myNameIsPatrick:
Well, I found the python 3.6 issue with the test. Mock open doesn't support reading in chunks, which was fixed for python 3.7+. See https://bugs.python.org/issue31855.

My solution is to just skip the test for this version of python for now, and when we figure out a better way of handling test files down the line, we can remove this condition.

myNameIsPatrick:
@bfc5288, thanks for the review. Most of the functionality is covered in `conftest.py` but it would be helpful to not bake in this functionality and use the test files directly. @shereenElSayed has explored this in the slack app using a pytest extension and it'll be good to extend that to the client library as well.

PR #55, opened 2020-03-25
Feature/gcn subscriber

bfc5288:
## Description

This PR implements #5, to be able to consume GCN from a broker notices by subscribing to a Kafka topic and print to stdout, e.g.:
```
scimma subscribe kafka://hostname:port/gcn -e
```

Supported command line options are:
* `-e` to read from the earliest broker offset (**required if one wants to consume previously-published messages**)
* `--json` to dump JSON to stdout, instead of printing the GCN
* `-t` to specify a timeout period to wait for new messages from the broker

Configuration options can be passed as with `scimma publish`.

Unit tests of the entrypoint and of the GCN printer have been added. A reference GCN raw stdout file has been added as the expected output for the GCN printer test (a regularly-formatted GCN file is not suitable for this task, which is sensitive to stdout nuances such as newline characters).

Documentation has been added to the README and quickstart.

## Checklist


doc/user/quickstart.rst
```diff
@@ -58,3 +58,15 @@ Client `configuration <https://github.com/edenhill/librdkafka/blob/master/CONFIG
...
+^^^^^^^^^^^^^
+
+.. code:: bash
+
+    scimma subscribe -b kafka://hostname:port/gcn mygcn.gcn3 -e
+
+Client `configuration <https://github.com/edenhill/librdkafka/blob/master/CONFIGURATION.md>`_
+properties can be passed to :code:`scimma subscribe` via :code:`-X property=value` or in a configuration
+file specified by :code:`-F <config-file>`, mimicking the behavior of :code:`kafkacat`. This can be
+used to connect to a Kafka broker with SSL authentication enabled, for example.
```
myNameIsPatrick: I would remove one of the paragraphs on the client configuration, probably the one with publishing GCNs to not duplicate it.

scimma/client/subscribe.py
```diff
@@ -0,0 +1,118 @@
+#!/usr/bin/env python
+
+__author__ = "Bryce Cousins (bfc5288@psu.edu)"
+__description__ = "tools to receive and parse GCN circulars"
+
+
+import argparse
+
+import json
```
myNameIsPatrick:
```suggestion
import argparse
import json
```

scimma/client/subscribe.py
```diff
@@ -0,0 +1,118 @@
...
+    start_offset = "latest"
+    timeout = 10
+    json_dump = False
+
+    if args.json:
+        json_dump = True
+    if args.earliest:
+        start_offset = "earliest"
+    if args.timeout:
+        timeout = int(args.timeout)
```
myNameIsPatrick: You can add a type to the argument option and won't need to cast this as a result.

scimma/client/subscribe.py
```diff
@@ -0,0 +1,118 @@
...
+    # consumer options
+    parser.add_argument(
+        "-j", "--json", help="Request gcn output as raw json", action="store_true",
+    )
+    parser.add_argument(
+        "-e",
+        "--earliest",
+        help="Request to stream from the earliest available Kafka offset",
+        action="store_true",
+    )
```
myNameIsPatrick:
```suggestion
    parser.add_argument(
        "-e",
        "--earliest",
        type=float,
        help="Request to stream from the earliest available Kafka offset",
        action="store_true",
    )
```

scimma/client/subscribe.py
```diff
@@ -0,0 +1,118 @@
...
+# -- main
+
+
+def _add_parser_args(parser):
+    parser.add_argument(
+        "-b",
+        "--broker-url",
+        required=True,
+        help="Sets the broker URL (kafka://host[:port]/topic) from which to receive GCNs.",
+    )
```
myNameIsPatrick: I know I caused this API incompatibility (my bad), but would it be possible to raise an issue about the inconsistency so we can address it later? Alternatively, we can change it in this PR depending on if you're willing:

```
    parser.add_argument(
        "url",
        metavar="URL",
        help="Sets the URL (kafka://host[:port]/topic) to publish GCNs to.",
    )
```

and then change the arguments appropriately in this module and in the tests.

tests/test_subscribe.py
```diff
@@ -0,0 +1,45 @@
...
+        gcn_stdout_str = "".join(gcn_stdout_list)
+
+        # read in expected data txt
+        shared_datadir = Path("tests/data")
+        expected_raw = (
+            shared_datadir / "expected_data" / "gcn_circular_stdout.txt"
+        ).read_text()
+
+        # use codec to decode the expected output to leave newline/tab characters intact
+        expected = codecs.unicode_escape_decode(expected_raw)[0][:-1]  # remove ending newline char
```
myNameIsPatrick: Seems kinda strange you have to do this, but I'm guessing it's because of the formatting of the file.

scimma/client/subscribe.py
bfc5288: Thanks, I was planning to add this update after completing the initial PR to be consistent with the #54 changes. See new commits on the PR for the changes.

tests/test_subscribe.py
bfc5288: Agreed, it is odd; it was the only way I could find at the time to read in the text exactly as it is written in the file (other methods like `readlines()`, `read()`, `read_text()`, etc. would decode the text and add escape backslashes `\` to all special characters. Options that I thought would avoid this, like `read_text(newline='')`, still had this behavior). 

I will keep an eye on this and update it if I find a better solution.

PR #57, opened 2020-03-31
Fix/dependencies-01

bfc5288:
## Description

This PR updates the Conda recipe build requirements, which had caused an error during the Conda package building/uploading workflow due to missing dependencies.

Due to the nature of the Conda build/upload workflow (the build workflow fails when pushing a release tag from a non-`master` branch, such as this `fix/` branch, due to a hash mismatch between the package built from the dev branch and the expected master repo hash), I replicated the necessary steps of the [workflow](https://github.com/scimma/client_library/edit/fix/dependencies-01/.github/workflows/upload_package.yml) locally:

* enter the PyPI package version at the top of `recipe/meta.yaml`:
    `{% set version = '0.0.4a4' %}`
* change the `source-sha256` hash to match that of the [PyPI package](https://pypi.org/project/scimma-client/0.0.4a4/#copy-hash-modal-38369969-35b3-4535-ae4b-60cb765ad7fe) in `recipe/meta.yaml`
    * (these first two steps are necessary to properly download from PyPI and build from the tarball without hash conflicts)
* add the SCIMMA and conda-forge channels for dependencies: 
`    conda config --env --add channels scimma  #for adc`
`    conda config --env --add channels conda-forge`

* build the package: `conda build .`

Following these steps on this development branch repo successfully built the package as expected.

## Checklist


recipe/meta.yaml
```diff
@@ -26,7 +26,9 @@ requirements:
     - setuptools_scm
   run:
     - python
-    - 'adc[channel=scimma]'
+    - adc >=0.0.3
```
myNameIsPatrick, line 29: is there a reason why the channel specifier was removed?

recipe/meta.yaml
```diff
@@ -26,7 +26,9 @@ requirements:
...
+    - xmltodict >=0.9.0
+    - dataclasses
```
myNameIsPatrick, line 31: We'll need to add a specifier here that says it's only needed for python 3.6. A guide to these are in here: https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#preprocessing-selectors

```suggestion
    - dataclasses # [py==36]
```

recipe/meta.yaml
bfc5288: Specifying both the channel and the version did not work during my tests (specifying the version requires quotes, which didn't appear to work when also specifying the version).

And it turns out that specifying the channel here is redundant since the `scimma` channel is already added in the [conda build/upload workflow](https://github.com/scimma/client_library/blob/ee6165e3274635ecda90573d44fb74170ae465cf/.github/workflows/upload_package.yml#L72), so `conda build` finds `adc` without specifying the channel here in the recipe.

bfc5288:
I made the recommended changes and tested that the build is skipped when expected, such as when running e.g. `conda build . --python=3.5`.

recipe/meta.yaml
myNameIsPatrick: Makes sense, thanks for the clarification.

PR #58, opened 2020-04-09
Migrate package name to hop

myNameIsPatrick:
## Description

Overview: Change package name to hop.

* imports: `import scimma.client` -> `import hop`
* repo: scimma/client_library -> scimma/hop
* package name: `scimma-client` -> `hop-client` on both PyPI/conda. It's not called `hop` due to package conflicts and looking at PyPI statistics on that package, there's still some download activity there (https://pypistats.org/packages/hop) so I think it may cause confusion if we try to request to use the `hop` name there.

Two other changes while doing this:
* Update the `genesis` import to use `adc-streaming` to comply with the new name change.
* Change how unit tests are run to be more robust since I was running into unit test failures due to imports.

## Checklist


myNameIsPatrick:
After this has been approved and merged in, I'll change the repo name appropriately and then work on transferring downstream repositories (such as the template repo) to follow suit.

recipe/meta.yaml
```diff
@@ -31,15 +31,14 @@ requirements:
 
 test:
   imports:
-    - scimma
-    - scimma.client
-    - scimma.client.publish
+    - hop
+    - hop.publish
```
bfc5288, line 36: I missed this when I added `scimma subscribe`, so we should include that test here:
```suggestion
    - hop.subscribe

```

setup.py
```diff
@@ -51,7 +50,7 @@
     extras_require = extras_require,
     setup_requires = ['setuptools_scm'],
```
bfc5288, line 51: Running `make test` failed for me:
```
$ make test
python setup.py test
usage: setup.py [global_opts] cmd1 [cmd1_opts] [cmd2 [cmd2_opts] ...]
   or: setup.py --help [cmd1 cmd2 ...]
   or: setup.py --help-commands
   or: setup.py cmd --help
error: invalid command 'pytest'
make: *** [Makefile:21: test] Error 1
```
It looks like `setup.py` is not set up to handle the `test` (equivalently `pytest`) argument. To do that, I believe you have to make these changes (refer to [here](https://doc.pytest.org/en/3.0.4/goodpractices.html#integrating-with-setuptools-python-setup-py-test-pytest-runner)):
```suggestion
    setup_requires = ['setuptools_scm', 'pytest-runner'],
    tests_require = ['pytest'],
```
After this change, `make test` worked for me.

doc/user/quickstart.rst
```diff
@@ -8,11 +8,11 @@ Quickstart
 Reading messages
 ----------------
 
-The scimma client supports a python-based API for reading messages from a stream, as follows:
+**hop** supports a python-based API for reading messages from a stream, as follows:
```
bfc5288, line 11: (see comment about **hop-client** left on the README)

README.md
```diff
@@ -1,52 +1,52 @@
-SCiMMA Client
+Hop
 =============
 
-![](https://github.com/scimma/client_library/workflows/build/badge.svg)
+![](https://github.com/scimma/hop/workflows/build/badge.svg)
 
-**scimma-client** is a pub-sub client library for Multimessenger Astrophysics.
+**hop** is a pub-sub client library for Multimessenger Astrophysics.
```
bfc5288, line 6: We may want to refer to the package in the user-facing docs as **hop-client** to match the name on the package repos (even though the code is restructured to be `hop/` rather than `hop/client/` -- thoughts?

doc/Makefile
```diff
@@ -4,7 +4,7 @@
 # You can set these variables from the command line.
 SPHINXOPTS    =
 SPHINXBUILD   = sphinx-build
-SPHINXPROJ    = scimma-client
+SPHINXPROJ    = hop
```
bfc5288, line 7: (see comment about **hop-client** left on the README)

doc/user/commands.rst
```diff
@@ -6,16 +6,16 @@ Commands
    :local:
 
 
-**scimma-client** provides a command line interface for various tasks:
+**hop** provides a command line interface for various tasks:
```
bfc5288, line 9: (see comment about **hop-client** left on the README)

doc/user/installation.rst
```diff
@@ -5,25 +5,25 @@ Installation
 .. contents::
    :local:
 
-You can install scimma-client either via pip, conda, or from source.
+You can install hop either via pip, conda, or from source.
```
bfc5288, line 8: (see comment about **hop-client** left on the README)

README.md
myNameIsPatrick: I think that sounds reasonable. I would also propose naming the repo scimma/hop-client to reflect this so it's all consistent.

setup.py
myNameIsPatrick: Nice catch, that's what I get for reusing the same environment (which had `pytest-runner` installed) for development.

myNameIsPatrick:
> @myNameIsPatrick Good work on all of the refactoring!
> 
> I couldn't run `make test` due to an issue with `setup.py` (see in-line comments). I posted a fix that made it work for me; let me know how this looks or if I missed something.

Thanks! I modified it a bit so that it's captured in the `dev` extra rather than installed by default but otherwise looks good. Tested on a fresh environment and didn't run into any issues.

> A general comment: we may want to refer to **hop** as **hop-client** in user-facing docs (to match what will be on the package repos), instead of **hop** as it is in the README and elsewhere (see in-lines).

Agreed, fixed those in the latest commit.

> After this, it would be good to push a tag to test the repo upload to ensure they accommodate the package name change.

Definitely.

myNameIsPatrick:
> Thanks, those changes look good.
> 
> Though I'm getting the same error as before when running `make test` on this version in a fresh environment, even if I add `tests_require = ['pytest']` as before or in the `extras_require` section. Are you sure that `pytest-runner` doesn't need to be in the `setup_requires` field?

Have you run `pip install .[dev]` first, or did you install the base package only?

bfc5288:
> > Thanks, those changes look good.
> > Though I'm getting the same error as before when running `make test` on this version in a fresh environment, even if I add `tests_require = ['pytest']` as before or in the `extras_require` section. Are you sure that `pytest-runner` doesn't need to be in the `setup_requires` field?
> 
> Have you run `pip install .[dev]` first, or did you install the base package only?

I hadn't. I just did this with a clean environment and it worked as expected. Approved for merge!

PR #60, opened 2020-04-10
Refactor common utilities in publish.py+subscribe.py

myNameIsPatrick:
## Description

Mostly tackling some refactoring after adding in the subscribe functionality so that we reuse code when applicable.

* Added a new class method in `models.GCNCircular` to make a `GCNCircular` from email-formatted text, to replace `publish.read_parse_gcn`.
* Defined a string representation for `GCNCircular` to produce pretty-formatted text, that can be used to print directly.
* Fixed some formatting issues with the circulars. It turns out the spacing between the subject headers and the content (DATE: <date>) isn't tab delimited, but left justified with some number of spaces. Adjusted the stdout expected text in the tests accordingly. If possible, I'd like confirmation that this looks reasonable for the subscribe functionality.
* Moved a few commonly used parser/config utilities in publish.py, subscribe.py to cli.py.
* Added some more submodules to the API reference.
* Moved some tests around to reflect this change. There are no cli/publish unit tests, but all of the functionality is covered in test_cli.py and improves test coverage by a few percent after the refactoring.
* Simplified some of the default argument setting in subscribe.py, the relevant defaults that are exposed are now listed when running `hop subscribe --help`.

## Checklist

* [ ] All new functions and classes are documented and adhere to Google doc style (3.8.3-3.8.6 of [this](http://google.github.io/styleguide/pyguide.html#383-functions-and-methods) document)
* [ ] Add/update sphinx documentation with any relevant changes.
* [ ] Add/update pytest-style tests in `/tests`, ensuring sufficient code coverage.
* [ ] `make test` runs without errors.
* [ ] `make lint` doesn't give any warnings.
* [ ] `make format` doesn't give any code formatting suggestions.
* [ ] `make doc` runs without errors and generated docs render correctly.
* [ ] Check that CI pipeline run on this PR passes all stages.
* [ ] Review signoff by at least one developer.

NOTE: If this PR relates to a release, open and reference an issue with the Release checklist template.

PR #68, opened 2020-05-13
Feature - add subscribe printing of notices

bfc5288:
## Description
This PR implements #25, namely a fix to the `hop-client` that allows `hop.subscribe` to correctly parse and print ADC stream messages that aren't formatted as GCN circulars. 

I also added the option for `subscribe` to never timeout, which is useful for letting a client stream messages from a topic indefinitely.

----

Design choices: I am letting `hop.io.models` contain the printing logic, such that `subscribe` is simply a parser that classifies a stream message as a GCNCircular/Notice/VOEvent dataclass and then calls that dataclass's print method (defined in `io.models`).

In this feature, messages are classified by substring-indexing an incoming message with a flag specific to a circular ("GCN Circular") or notice ("ivorn"). As @spenczar noted, this is not a long-term solution for message classification, which will require reconsideration of the message source (e.g., Kafka topics and brokers) and message content (e.g., circular, VOEvent).

----

With these changes, a VOEvent message from an ADC stream prints as:

<details>
<summary>hop.subscribe VOEvent stdout</summary>

```
##################################################
######## Hop-client: parsing a VOEvent  ##########
##################################################
{
  "ivorn": "ivo://nasa.gsfc.gcn/SWIFT#BAT_Known_Pos_-1855369095-174",
  "role": "observation",
  "version": "2.0",
  "Who": {
    "AuthorIVORN": "ivo://nasa.gsfc.tan/gcn",
    "Author": {
      "shortName": "VO-GCN",
      "contactName": "Scott Barthelmy",
      "contactPhone": "+1-301-286-3106",
      "contactEmail": "scott.barthelmy@nasa.gov"
    },
    "Date": "2020-05-01T09:55:00",
    "Description": "This VOEvent message was created with GCN VOE version: 1.47 08mar20"
  },
  "What": {
    "Param": [
      {
        "name": "Packet_Type",
        "value": "141"
      },
   ........SNIP........
  },
  "WhereWhen": {
    "ObsDataLocation": {
      "ObservatoryLocation": {
        "id": "GEOLUN"
      },
   ........SNIP........
  },
  "How": {
    "Description": "Swift Satellite, BAT Instrument",
    "Reference": {
      "uri": "http://gcn.gsfc.nasa.gov/swift.html",
      "type": "url"
    }
  },
  "Why": {
    "importance": "0.90",
    "Inference": {
      "probability": "0.98",
      "Name": "Crab",
      "Concept": "process.variation.burst;em.gamma"
    }
  },
  "Citations": {},
  "Description": null,
  "Reference": {}
}
```
</details>
This is just the json dump of the VOEvent, so it is a bit messy. We should look into how to better print VOEvent and xml-formatted messages in the future.

## Checklist


hop/subscribe.py
```diff
@@ -51,9 +84,9 @@ def _add_parser_args(parser):
     parser.add_argument(
         "-t",
         "--timeout",
-        type=float,
         default=10,
-        help="Specifies the time (in seconds) to wait for new messages. Default: 10s",
+        help="Specifies the time (in seconds) to wait for messages before timing out; "
+        "specify 'None' to wait indefinitely.  Default: 10 seconds",
```
spenczar, line 89: I think a more common pattern in unix-y tools would be to specify a negative value to wait indefinitely.

hop/subscribe.py
```diff
@@ -72,12 +105,16 @@ def _main(args=None):
     # set offset
     start_offset = "earliest" if args.earliest else "latest"
 
+    # set timeout
+    t = args.timeout
```
spenczar, line 109: I'm a little wary of the way this code is shaping up. As it starts to use attributes on the `args` object, it gets harder to read. To know what `args.timeout` is, you have to read back upwards to get to `args = parser.parse_args`... but you also have to be prepared for `args` to be passed in to `_main` directly, and cross your fingers that it's an object with the right attributes. Meanwhile, they're also `config` which loads more configuration.

I think this PR represents the last limits before this really gets too unwieldy to keep working with it. We should probably refactor how configuration loading works before going any further - that can be a follow-on ticket.

hop/subscribe.py
```diff
@@ -6,28 +6,61 @@
...
-
     """
+
+    # check for msg format using standard-specific flags
+    voevent_flag = 'ivorn'
+    gcncir_flag = 'GCN CIRCULAR'
+
+    # VOEvent:
+    if voevent_flag in msg:
+        gcn = VOEvent(**msg)
```
spenczar, line 34: Please rename the `gcn` variable now that it can be a VOEvent message.

hop/subscribe.py
```diff
@@ -6,28 +6,61 @@
...
-      gcn_dict:  the dictionary object containing a GCN-formatted message
+      msg:       the raw message from an ADC stream
+      json_dump: boolean indicating whether to print msg as raw json
 
     Returns:
       None
-
     """
+
+    # check for msg format using standard-specific flags
```
spenczar, line 28: Oh man, this looks hard to maintain and destined to cause bugs: we're doing a string substring match to tell what type of message it is. This will take `O(n*m)` for `n` message formats of `m` length, which we will certainly regret eventually. Can we do something better? @adambrazier this has important architectural consequences: how do we discriminate between message types?

We should know based on the Kafka topic, but this code doesn't know what topic we're listening to at all - that's all abstracted away by `adc-streaming`, since we want to be agnostic to Kafka. For the same reason, we also can't use Kafka message metadata.

As I see it, that only leaves making a wrapper schema around any messages. Something like `{"type": "GCN Circular", "payload": {....}}`. We'd have to enforce that schema on the publish side probably. There are still choices to make: is JSON the right wrapper language, for example? The modern "right" answer in Kafka is almost certainly Avro.

It's fair to table this for after the workshop, but we really need to figure this out I think.

hop/subscribe.py
bfc5288: Agreed, that was my first thought -- I went with `None` since that was the convention in ADC and ConfluentKafka, though I could change this option on the user's side.

hop/subscribe.py
bfc5288: Sure, we can look into this more during a refactor. I based this setup on the [cli changes](https://github.com/scimma/hop-client/pull/60/files#diff-3a67069fe9c758c5cdc533b4d7ac90b3) from @myNameIsPatrick but it'd be good to consider the downstream effects of this.

hop/subscribe.py
bfc5288: I believe the nomenclature has `gcn` as the parent message class, where `gcn circular` and `gcn notice` (which includes `VOEvent`) are instantiations of `gcn`.

hop/subscribe.py
bfc5288: I agree; I added it as a purely temporary placeholder since I'm certain there's a better way to do this with more time and by clarifying how we want to handle message type (at both the consumer and producer levels).

You're right -- the ultimate solution shouldn't rely on kafka topic, let alone message strings. There's more to think about for the right implementation (time for a user story), though for this release the substring holdover works for the AWS kafka instance.

bfc5288:
> Some early comments, even though this is a draft.

Thank you @spenczar! Those are very good points -- I left comments, but am not able to resolve some of the core issues for the workshop (refactoring and message typing -- it looks like we have some important architecture user stories for next time!).

hop/subscribe.py
myNameIsPatrick: @spenczar, I can hopefully give some context into some of this.

So this whole business with `_main` and handling the `args` case is probably not needed, it's a pattern I've normally followed when building a CLI with a subparser so that subcommands are registered correctly, via https://github.com/scimma/hop-client/blob/feature/subscribe_notices/hop/__main__.py#L43. With some work, it can probably be simplified.

The rationale behind the `config` handling was to provide a way to do auth before some of the auth functionality was added to `adc-streaming`, so the goal there was to mimic how config handling is done via kafkacat (the -F and -X opts) to pass in directly to the stream instance. I would prefer some simplification of this and to provide a simpler auth interface and some defaults so that it "does the right thing" connecting to the hopskotch broker, and axe the custom config stuff entirely. That may or may not be doable, but having flexibility like this comes with additional complexity.

We should figure out what it is we ideally want to expose and then refactor our CLI as needed to provide a simple, consistent interface that does what we want. There's definitely some technical debt we'd need to pay off as we've added in features one by one over the sprints.

hop/models.py
```diff
@@ -82,7 +86,9 @@ def asdict(self):
         return asdict(self)
 
     def __str__(self):
-        headers = [(name.upper() + ":").ljust(9) + val for name, val in self.header.items()]
+        headers = [
```
myNameIsPatrick, line 89: This is out of scope for this PR, but I would be inclined to pick up the `autopep8` with associated config as @spenczar added for `adc-streaming` as a future task, as we can allow the 100 line limit in our 'format guide' and avoid formatting stuff like this.

hop/subscribe.py
myNameIsPatrick: I'm inclined to agree with renaming the `gcn` variable as well, maybe something like `model`(?) to reflect the return description in the docstring, "dataclass model object for the raw message".

hop/subscribe.py
myNameIsPatrick: > It's fair to table this for after the workshop, but we really need to figure this out I think.

Definitely. If we have guarantees for what messages come out of certain topics, then we don't need message metadata to determine the message type. What we're lacking at the moment is a specification for our API/schema/serialization, however minimal that may be to be able to answer questions like these.

hop/subscribe.py
```diff
@@ -6,28 +6,69 @@
...
+            gcn = GCNCircular(**msg)
+            status_str = "## Parsing a hop-published message"
+        except:
+            warnings.warn("## Warning: message format not recognized; dumping as json")
+            gcn = msg
+            status_str = "## Dumping unknown message as json"
+
+            print(status_str)
+            print_gcn(gcn, json_dump=True)
+            return None
```
myNameIsPatrick, line 50: What is the rationale behind the second try/except? I'd imagine is it isn't recognized as falling into a VOEvent or GCNCircular above, then it should just be treated as a message whose type that can't be recognized automatically, but maybe I'm missing something.

I'm not sure how I feel about the warning, but I don't think we discussed whether the hop client should or should not handle arbitrary message formats.

hop/subscribe.py
```diff
@@ -72,12 +113,20 @@ def _main(args=None):
     # set offset
     start_offset = "earliest" if args.earliest else "latest"
 
+    # set timeout
+    t = args.timeout
+    timeout = None if t == "-1" else float(t)
```
myNameIsPatrick, line 118: Why was the `type=float` option removed from the timeout arg? I don't think you'll need to do the float casting or handling of -1 as a string with that in place.

hop/subscribe.py
```diff
@@ -72,12 +113,20 @@ def _main(args=None):
...
     stream = Stream(format=gcn_format, config=config, start_at=start_offset)
     with stream.open(args.url, "r") as s:
-        for gcn_dict in s(timeout=args.timeout):
-            print_gcn(gcn_dict, args.json)
+        for msg in s(timeout=timeout):
+            gcn = classify_msg(msg)
+            if gcn is not None:
+                print_gcn(gcn, args.json)
+            else:
+                continue
```
myNameIsPatrick, line 132: I don't think you need this else statement.

hop/subscribe.py
bfc5288: Good catch -- I removed `type` so that the user could specify `None` for indefinite timeout, but changed this to `-1`. Adding `type` back in.

hop/subscribe.py
bfc5288: I added the `except` to handle unknown input (like a string instead of a file) during testing, though I think `hop` should fail in that case rather than attempt printing. I'll remove this. Later we should plan to figure out whether hop-client should handle arbitrary input (along with message typing itself).

hop/subscribe.py
bfc5288: That's right, thanks. Removed.

myNameIsPatrick:
@bfc5288, feel free to resolve any conversations you addressed. As an aside, do you know how to retrigger that codecov report? Although it's unclear, it seems like the report is being updated. I remember the code coverage was much lower than it is in your latest commit.

Anywho, if there's a drop in the coverage due to additions that aren't being addressed in the testing, I would have a preference to either address those extra cases in this PR or to open up an issue so we can tackle them after the tutorial sesssion since this is getting a bit time-critical.

bfc5288:
> @bfc5288, feel free to resolve any conversations you addressed. As an aside, do you know how to retrigger that codecov report? Although it's unclear, it seems like the report is being updated. I remember the code coverage was much lower than it is in your latest commit.
It looks like the Codecov report has been rerunning, but the plugin hasn't re-commented the new reports. This would have to be changed in the Codecov workflow somewhere.

> Anywho, if there's a drop in the coverage due to additions that aren't being addressed in the testing, I would have a preference to either address those extra cases in this PR or to open up an issue so we can tackle them after the tutorial sesssion since this is getting a bit time-critical.

Good idea -- the large drop in coverage is due to the `classify_msg` in subscribe, which will be retooled anyways when we determine the best way to handle message typing. I could set up some basic tests to make sure the substring indexing works as expected for VOEvents and circulars, though this may be quickly deprecated since we'll be be changing to another method to classify messages. Opening an issue would probably be best.

myNameIsPatrick:
> It looks like the Codecov report has been rerunning, but the plugin hasn't re-commented the new reports. This would have to be changed in the Codecov workflow somewhere.

Sounds reasonable, I wouldn't worry about it here but something to keep in mind.

> Good idea -- the large drop in coverage is due to the `classify_msg` in subscribe, which will be retooled anyways when we determine the best way to handle message typing. I could set up some basic tests to make sure the substring indexing works as expected for VOEvents and circulars, though this may be quickly deprecated since we'll be be changing to another method to classify messages. Opening an issue would probably be best.

Agreed, let's punt the issue when we figure out a better solution to determining the message type.

doc/user/commands.rst
```diff
@@ -8,14 +8,25 @@ Commands
...
-
+:code:`hop publish`: parse and publish GCN circulars and notices
+:code:`hop subscribe`: parse and print GCN circulars and notices
 
 :code:`hop publish`
 ~~~~~~~~~~~~~~~~~~~~~~
 
-This command allows a user to parse an RFC 833 formatted GCN circular and
-publish JSON-formatted GCNs via Kafka.
+This command allows a user to parse an `RFC 833 formatted GCN circular <https://gcn.gsfc.nasa.gov/gcn3_circulars.html>`_
```
spenczar, line 17: Below it's described as RFC 822. Which one is right? I'm not sure.

myNameIsPatrick: That's a typo, it should be RFC 822.

Issue #75, opened 2020-06-12
UX: Add an option to the library + cmdline client to upload opaque blobs.


myNameIsPatrick:
@patrick

myNameIsPatrick:
@mjuric, I'm looking into this now and I'm wondering how this is different from https://github.com/scimma/architecture/issues/31. It seems like this issue is a subset of the one I linked here.

myNameIsPatrick:
This is essentially a subset of https://github.com/scimma/architecture/issues/31, so I'll be marking this as done.

PR #80, opened 2020-06-19
Feature/message wrapper

bfc5288:
## Description

This PR implements https://github.com/scimma/architecture/issues/31, namely to restructure messages so that they're wrapped as:
```
{
    "format ": "message_format"
    "content": " message_content"
}
```

This allows `hop subscribe` to more reliably classify incoming messages without dubious substring parsing (as in #68). This relies on the publisher to specify their message format during the send; otherwise the message is sent as an unstructured blob which `subscribe` handles in turn (hence implementing #75).

Currently, `hop publish` supports publishing multiple messages if they're of the same format via:
```
hop publish URL -f <file-format> file1 file2 file3 ....
```
An alternative would be to allow the publisher to send multiple messages of different formats, e.g.:
```
hop publish URL -f format1 file1 -f format2 file2 ....
```
which could be implemented with a few changes to the publish CLI.

To prepare to generalize to future message formats, I also renamed the `gcn` format to `circular` and ensured that `gcn` wasn't used to refer to general message variables (`message` is used instead).

I added unit tests for the new functionality and modified existing tests when appropriate, though there are a few changes we could still consider (e.g., parametrizing tests in `test_cli` to reduce repeated code [see test_subscribe]; potentially moving more towards the `expected_data`/`test_data` structure instead of using definitions in `conftest`, etc.).

## Checklist


hop/publish.py
```diff
@@ -19,37 +20,54 @@
...
+                with open(message_file, "rb") as f:
+                    message_model = VOEvent.from_xml(f)
+            elif message_format == "blob":
+                with open(message_file, "r") as f:
+                    message_model = message_blob.from_file(f)
             else:
-                raise ValueError(f"File extension {extension} not recognized")
+                warnings.warn("Warning: format not recognized. Sending as unstructured blob")
+                with open(message_file, "r") as f:
+                    message_model = message_blob.from_file(f)
```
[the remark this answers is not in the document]
spenczar: One way to get around the `"rb"` vs `"r"` file mode differences is for each function in the `model_loader` map to take a filename, so like 
```python
formats = {
    "circular": GCNCircular.from_email_file,
    "voevent": VOEvent.from_xml_file,
    "blob": message_blob.from_file,
}
...
if message_format in formats:
    loader = formats[message_format]
else:
    # raise the warning, etc
for message_file in args.message:
    message_model = loader(message_file)
```
```python
class VOEvent:
   ...
    @classmethod
    def from_xml_file(cls, filename):
        with open(filename, "rb") as f:
            return cls.from_xml(f)
```

hop/subscribe.py
```diff
@@ -1,64 +1,64 @@
...
-        status_str = "## Parsing a GCN Circular"
-    else:
-        msg_model = GCNCircular(**msg)
-        status_str = "## Parsing a hop-published message"
 
-    print(status_str)
+    try:
+        fmt = message["format"]
+        content = message["content"]
+    except TypeError:
```
spenczar, line 30: I think also want to check for `KeyError`, in case `message` is a dict but it just doesn't have those keys.

hop/publish.py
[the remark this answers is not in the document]
bfc5288: Thanks for the pointers. I implemented the mapping with @spenczar's structure so that the file loading can now happen from `models.py`. (I also used the mapping structure in a similar setup to generalize the unit tests)

hop/subscribe.py
bfc5288: Good catch, I added this and a test for both errors.

bfc5288:
> Looks good, @bfc5288! I added some review comments/suggestions.
Thanks for the comments! I added revisions.

For unit test restructuring, what seemed most general was to parametrize each test with a list of message formats. Then, the test accesses a nested dictionary fixture from `conftest`, which holds the parameters for each message type (name, model, file, and text).

This gets around being unable to have fixtures in parametrizations (e.g., parametrizing test_cli_publish with the `voevent_text` fixture), and allows `test_cli` and `test_subscribe` to be easily extended for future message types (i.e., putting message name in the parametrization, and its parameters in `conftest`). For now, `test_models` probably still has to have separate test functions for each message type.

I'm wrapping up the changes and will update the docs next.

hop/models.py
```diff
@@ -103,12 +139,79 @@ def from_email(cls, email_input):
...
         else:
-            msg = email.message_from_string(email_input)
+            message = email.message_from_string(email_input)
 
         # format gcn circular into header/body
         return cls(
-            header={title.lower(): content for title, content in msg.items()},
-            body=msg.get_payload(),
+            header={title.lower(): content for title,
+                    content in message.items()},
```
myNameIsPatrick, line 149: Is there any way to get this to fit into one line instead? It's harder to see that it's a dict comprehension when it's split like this.

hop/subscribe.py
```diff
@@ -1,64 +1,67 @@
...
+        raise ValueError("Message is not wrapped with format/content keys")
+    except KeyError:
+        raise KeyError("Message does not contain format/content keys")
+
+    # create the dataclass model appropriate for the message format
+    model_creator = {"circular": GCNCircular,
+                     "voevent": VOEvent,
+                     "blob": MessageBlob,
+                     }
```
myNameIsPatrick, line 39: Same formatting as above if applicable.

hop/publish.py
```diff
@@ -38,18 +48,19 @@ def _main(args=None):
...
-            else:
-                raise ValueError(f"File extension {extension} not recognized")
-
-            # publish GCN
-            s.write(gcn.asdict())
+
+        model_loader = {"circular": GCNCircular.from_email_file,
+                        "voevent": VOEvent.from_xml_file,
+                        "blob": MessageBlob.from_text,
+                        }
```
myNameIsPatrick, line 55: I think it may be cleaner if we move this outside of the stream context. Also, I think it may be nicer if it's formatted as such:
```
model_loader = {
    "circular": GCNCircular.from_email_file,
    "voevent": VOEvent.from_xml_file,
    "blob": MessageBlob.from_text,
}
```

PEP8 is a little more loose on this requirement but there's already precedent for this formatting in the argparse arguments and I'd like to be consistent.

tests/conftest.py
```diff
@@ -193,3 +196,43 @@ def voevent_fileobj():
...
+                        "expected_model": models.VOEvent,
+                        "test_file": "example_voevent.xml",
+                        "model_text": VOEVENT_XML.encode(),
+                        },
+            "blob":     {"model_name": "MessageBlob",
+                         "expected_model": models.MessageBlob,
+                         "test_file": "example_blob.txt",
+                         "model_text": MESSAGE_BLOB,
+                         },
+            }
```
myNameIsPatrick, line 238: Same formatting as above.

hop/models.py
bfc5288: Yep, good catch. That was an auto-format change which definitely makes it less readable. Fixed.

Issue #74, opened 2020-06-12
UX: Add central configuration file support for the library/cmdline client.


myNameIsPatrick:
@patrick

PR #82, opened 2020-06-29
API overhaul in moving to adc-streaming 1.x

myNameIsPatrick:
## Description

This PR implements a number of changes to adapt to adc-streaming 1.x, as well as additional backwards-incompatible API changes. Since this needed to be done anyways for the upstream package, I took the liberty of making simplifications throughout as needed. This also implements https://github.com/scimma/hop-client/issues/74 as it was easier wrapping this all up as part of these changes.

#### Overview:

* Add new `hop auth` command with various utilities to create auth configuration and locate the path where this configuration is stored, used in `hop publish` and `hop subscribe` to authenticate.
* Simplify the `hop publish` and `hop subscribe` API substantially, removing the `-F` and `-X` commands to pass in arbitrary configuration variables. Instead. this is all handled by the auth configuration above and loaded in by default, unless `--no-auth` is specified. There is also no `--timeout` arg for `hop subscribe`, instead it reads messages until EOS is received. In order to open a long-lived connection, you can pass in `--persist` instead.
* Move all the functionality from `subscribe.classify_message` to a `Deserializer` class in `hop.io`. This is so the wrapping + serialization (or unwrapping+deserialization) of messages happens within the client itself.
* Standardize all of the message dataclass methods so that there's consistency between them.
* Removed this business with `if not args:` in the entry points since it's basically unused and adds needless complexity.
* Rewrite many of the docs in general to expose the various functionality. Some of the docs were also stale/incorrect and so I fixed those up as needed as well.
* Add some unit tests throughout to bring the test coverage above 90% again.

#### Selected files and motivation for changes:

Makefile:

* Bumped up the complexity of `flake8` to 12 for raising warnings. This is because `io.Stream.open()` was raising warnings, and it was a bit obnoxious.

README.md:

* Put all badges on one line
* Put a link to stable version of RTD rocs

hop/__init__.py:

* remove the silent `ImportError` behavior if `_version.py` is not found, which was probably more trouble than its worth and not easily testable.

hop/setup.cfg:

* Ignore unused local variables in the unit tests. I'm inclined to leave these in for clarity even if sometimes they aren't used.

hop/auth.py:

* Expose various authentication classes from adc-streaming
* Add CLI utilities. `hop auth locate` shows where the default auth config location lives, and can be controlled by $XDG_CONFIG_HOME. It uses `get_auth_path()` to find this location. `hop auth setup` creates an auth config file in that location if it doesn't already exist, and warns the user and does nothing if it does exist. You can overwrite the existing file with `hop auth setup --force`.
* `load_auth()` loads auth config from that default location and creates an `Auth` instead, to be passed into a `stream` instance.

hop/cli.py:
* Rename `add_url_opts` to `add_client_opts` since this also now includes the `--no-auth` option to be passed in by both publish/subscribe commands.

hop/io.py:
* Overhaul the streaming functionality for changes to adc-streaming. The easiest way I found to expose most of the same functionality as before was to subclass the `Producer` and `Consumer` classes and redefine how messages are serialized and deserialized to do message packing and to return the right message classes seamlessly. A lot of the functionality of `subscribe.classify_message` has been folded into the `Deserializer` class, which is used in the consumer. Most of these extra functions/classes, including exposing metadata is private since I don't want to expose any of this to the user.
* Expose the `ConsumerStartPosition` from adc-streaming to control the message offset while not having to call adc directly.

hop/models.py:
* Add feature parity to all the message dataclasses, renaming all the loading functionality to `load()` and `load_file()`, which should hopefully be more straightforward to handle message loading in the same way across all of these.
* Rename `wrap_message()` to `serialize()` since this may actually serialize in the future and provide some future-proofing.

hop/publish.py:
* Control the `--format` argument options with the members of `io.Deserializer`. This is so users are required to choose between these options.

hop/subscribe.py:
* Swap `--earliest` with a `--start-at`, with the same enum control as with `--format` above. To specify earliest, one can do `--start-at EARLIEST` now.
* Remove `--timeout`. By default, this will listen until EOS. If a long-lived connection is desired, the `--persist` command is exposed.

#### Extra notes

* While looking at the checklist below, I may have some indentation stuff to fix relating to the docstrings which I'll be fixing later today. (now fixed)

## Checklist


hop/io.py
```diff
@@ -3,44 +3,218 @@
...
+                _Metadata(
+                    message.topic(),
+                    message.partition(),
+                    message.offset(),
+                    message.timestamp()[1],
+                    message.key(),
+                )
+            )
+        else:
+            return payload
```
spenczar, line 183: How does a user of this API commit their message offset read position?

hop/auth.py
```diff
@@ -0,0 +1,124 @@
...
+        password = config["sasl.password"]
+        method = SASLMethod[config["sasl.mechanism"].replace("-", "_")]
+    except KeyError:
+        raise KeyError("configuration file is not configured correctly")
+    else:
+        return Auth(user, password, ssl=ssl, method=method, **extra_kwargs)
+
+
+# ------------------------------------------------
+# -- main
```
spenczar, line 86: cruft?

hop/auth.py
```diff
@@ -0,0 +1,124 @@
...
+import logging
+import os
+
+from adc import auth
+
+DEFAULT_AUTH_CONFIG = """\
+security.protocol=SASL_SSL
+sasl.mechanism=PLAIN
+sasl.username=username
+sasl.password=password
```
spenczar, line 17: I think we should consider designing a different configuration file format. This is the `librdkafka` INI config, and it's kind of inscrutable, and filled with options which our users shouldn't care need to set like `sasl.mechanism` and `security.protocol`.

I'd be happier with a `hop/config.toml` file that looks like
```toml
[auth]
username = "swnelson"
password = "..."
```

which also leaves room for more configuration in the future, eg 
```toml
[aliases.dev-gcn]
broker = "dev.hop.scimma.org"
topic = "gcn"
consume_group_id = "swnelson_group"
publish_format = "CIRCULAR"
```
(just as a totally-made-up example - the particulars dont make sense, but you get the idea).

hop/auth.py
```diff
@@ -0,0 +1,124 @@
...
+        # SSL options
+        if config["security.protocol"] == "SASL_PLAINTEXT":
+            ssl = False
+        elif "ssl.ca.location" in config:
+            extra_kwargs["ssl_ca_location"] = config["ssl.ca.location"]
+
+        # SASL options
+        user = config["sasl.username"]
+        password = config["sasl.password"]
+        method = SASLMethod[config["sasl.mechanism"].replace("-", "_")]
```
spenczar, line 78: We will likely only want to support `SHA-512` SCRAM auth. It might just be an error to provide anything else, soon. 

https://github.com/scimma/scimma-admin/blob/302b95d5f7d0468e0d4907231fd4f072b6d04352/scimma_admin/hopskotch_auth/models.py#L176

We could add that check after the scimma-admin auth mechanisms stabilize, though.

hop/auth.py
```diff
@@ -0,0 +1,124 @@
...
+import argparse
+import errno
+import logging
+import os
+
+from adc import auth
+
+DEFAULT_AUTH_CONFIG = """\
+security.protocol=SASL_SSL
+sasl.mechanism=PLAIN
```
spenczar, line 15: We should default to `SCRAM-SHA-512` here.

tests/test_io.py
```diff
@@ -4,29 +4,141 @@
...
+
+
+def test_pack(circular_msg, circular_text):
+    # message class
+    circular = GCNCircular(**circular_msg)
+    packed_msg = io._Producer.pack(circular)
+
+    # unstructured message
+    message = {"hey": "you"}
+    packed = io._Producer.pack(message)
```
spenczar, line 144: Maybe add a round-trip test for each format: pack a message and then unpack it; it should be unchanged.

tests/test_io.py
```diff
@@ -4,29 +4,141 @@
...
+
+        stream = io.Stream(persist=persist)
+
+        with stream.open(broker_url1, "r", start_at=start_at) as s:
+            for msg in s:
+                continue
+
+        with stream.open(broker_url2, "r", start_at=start_at) as s:
+            for msg in s:
+                continue
```
spenczar, line 90: What does this test actually test? I think all that's _really_ getting tested is that `stream.open` does not raise, and that it returns the contents of `mock_consumer.stream` (in other words, that you're mocking works. One issue with using mocks is that you can wind up testing that your mocking library is working, but not whether your code is correct.

Maybe that's sufficient here, but it's a little fishy to me - it doesn't test that `start_at` or `persist` or the URL are handled correctly, just that these options don't raise an exception.

Some ways you could improve this situation:
 - You could write a complicated mock. But I think you'll just be testing your mock code, as I said.
 - You could write integration tests which actually start up a Kafka broker in a container. But these tests can take a long time to execute and there's a lot of complexity in setting them up.
 - You could refactor code until it's unit-testable without mocks. For example, you might break up the `open` function in `io.py` to an `_open_reader` and `_open_writer`; your test of `open` is just whether it calls the right function based on the `mode` argument, and then you thoroughly test each of those new functions. But it's not clear whether this is actually an improvement to the code, other than making it more unit-testable. Obsessive dedication to testability isn't always right!
 - You could just keep the code as-is, relatively lightly tested, relying on your `test_cli` high-level tests rather than expecting unit tests to do much for you. This is probably what I would do.

Anyway, that's a long comment just to end up at "nothing to do here" but I think it's an interesting design question, and you might want to take one of those other paths.

README.md
```diff
@@ -2,31 +2,41 @@ Hop Client
...
 
-Client [configuration](https://github.com/edenhill/librdkafka/blob/master/CONFIGURATION.md)
-properties can be passed to `hop publish` via `-X property=value` or in a configuration
-file specified by `-F <config-file>`, mimicking the behavior of `kafkacat`. This can be
-used to connect to a Kafka broker with SSL authentication enabled, for example.
+By default, authentication is enabled, reading in configuration settings
+from `auth.conf`. The path to this configuration can be found by running
+`hop auth locate`. One can initialize this configuration with default
+settings by running `hop auth setup`. To disable authentication in the CLI
+client, one can run `--no-auth`.
```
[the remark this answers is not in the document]
myNameIsPatrick: That's a good idea, I have changed the order of this (and in the RTD quickstart) for this paragraph to go first.

doc/user/quickstart.rst
```diff
@@ -9,63 +9,115 @@ Quickstart
...
+settings by running :code:`hop auth setup`. To disable authentication in the CLI
+client, one can run :code:`--no-auth`.
 
-Reading messages from stream
-----------------------------
+Using the Python API
+----------------------
 
-The hop client supports a python-based API for reading messages from a stream, as follows:
+Publish messages
```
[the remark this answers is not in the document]
myNameIsPatrick: Nice catch, I've done a little more rearranging so that the order of producing/consuming messages is the same for both the CLI and python API sections.

hop/io.py
```diff
@@ -3,44 +3,218 @@
 __author__ = "Patrick Godwin (patrick.godwin@psu.edu)"
 __description__ = "a module for i/o utilities"
 
+from collections import namedtuple
```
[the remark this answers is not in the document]
myNameIsPatrick: I can do a dataclass. I normally lean towards namedtuples if the classes are simple (not too many attributes, no extra methods), but it also lacks a good way of adding docstrings which are nice.

hop/io.py
```diff
@@ -3,44 +3,218 @@
...
+                warnings.warn("group ID has no effect when opening a stream in write mode")
+            if start_at is not None:
+                warnings.warn("start_at has no effect when opening a stream in write mode")
+            if persist is not None:
+                warnings.warn("read_forever has no effect when opening a stream in write mode")
+            return _open_producer(broker_addresses, topics[0], auth=auth)
+        elif mode == "r":
+            if group_id is None:
+                group_id = _generate_group_id(10)
+                logger.info("group ID not specified, generating a random group ID")
```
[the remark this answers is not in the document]
myNameIsPatrick: Done.

hop/io.py
```diff
@@ -3,44 +3,218 @@
...
         Kwargs:
-          mode: read ('r') or write ('w') from the stream
-          format: the message serialization format
-          start_at: the message offset to start at
-          config: librdkafka style options, either a `dict` or a file path
+            mode: Read ('r') or write ('w') from the stream.
+            auth: An Auth instance.
+            start_at: The message offset to start at (read only).
+            persist: Whether to listen to new messages forever or stop
+                 when EOS is received (read only).
```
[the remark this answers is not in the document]
myNameIsPatrick: I can give some motivation for why I took this approach, and then you can tell me whether or not this is a crazy idea. So, I was trying to emulate request's `Session` object (https://requests.readthedocs.io/en/master/user/advanced/) here when making the `Stream` here to persist certain parameters across open connections, but being able to override them as needed when making a call to `open()`. Part of this is a bit silly since we're not really keeping around the same Producer/Consumer objects across multiple calls to `open()`, so some of the benefits of keeping this doesn't really apply here. Nonetheless, I was hoping that some of this was handy even if it meant keeping a little extra complexity on our side. In particular, I wanted to support two ends of use cases while making it kinda nice and flexible to do both. 

In the first, I only care about making a single connection and don't care about just putting everything into the `open()` call. In that case, carrying around the burden of the `Stream` class is a little much and I added a convenience wrapper, `stream`, which maps directly to `Stream()`. As a result, I can do something like this:

```
from hop import stream

with stream.open("kafka://myhostname/topic", "r", persist=True) as s:
    for message in s:
        print(s)
```

In the second case, I would opening up multiple producers and/or consumers and may want to save some common configuration ahead of time, such as providing auth credentials. Say, reading from a topic and writing the same message to another topic. Then I could do something like:

```
from hop import Stream
from hop.auth import load_auth

s = Stream(auth=load_auth(), persist=True)
with s.open("kafka://myhostname/topic1", "r") as source, s.open("kafka://myhostname/topic2", "w") as sink:
    for message in source:
        sink.write(message)
```

hop/io.py
```diff
@@ -3,44 +3,218 @@
...
+        try:
+            format = message["format"].upper()
+            content = message["content"]
+        except (KeyError, TypeError):
+            raise ValueError("Message is incorrectly formatted")
+        else:
+            if format in cls.__members__:
+                return cls[format.upper()].value(**content)
+            else:
+                raise ValueError(f"Message format {format} not recognized")
```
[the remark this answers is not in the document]
myNameIsPatrick: I think that's a good idea. What do you imagine we should expose to the `MessageBlob` to make this clear? Or do you think a log message (maybe at log level warning) would be more suitable?

hop/io.py
```diff
@@ -3,44 +3,218 @@
...
+                **opts,
+            )
+        else:
+            raise ValueError("mode must be either 'w' or 'r'")
+
+
+class Deserializer(Enum):
+    CIRCULAR = models.GCNCircular
+    VOEVENT = models.VOEvent
+    BLOB = models.MessageBlob
```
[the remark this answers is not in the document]
myNameIsPatrick: I'm really keen on the plugin approach and was thinking about that myself as well. Others could define new message types and as long as it's registered and defines the right behavior, it can be handled in the same way as the base classes. I think it may be worth it for the reasons you mentioned, despite the extra complexity.

I've been personally eyeing the pluggy library (https://pluggy.readthedocs.io/en/latest/) for supporting a plugin architecture for another project and was wondering your thoughts on it versus rolling our own implementation. It's pretty minimal (1 dependency if python < 3.8) and used to be part of pytest before they split it off into their own thing.

hop/io.py
myNameIsPatrick: In short, they don't. That wasn't supported before either, but I don't see a reason why we couldn't support that now. In principle, that shouldn't be hard to do by overriding `adc.Consumer.mark_done()` to take in a `Metadata` object instead of `confluent_kafka.Message()` since it can build a `TopicPartition` from that and store the offset that way.

hop/auth.py
myNameIsPatrick: Old habits die hard, man. That's also present in `publish.py` and `subscribe.py`, and it's something I have done to separate sections. Maybe we should just axe all these comment sections and remove this `__author__` and `__description__` stuff that we have all over (can blame me for this also).

hop/auth.py
myNameIsPatrick: I agree wholeheartedly with this. I don't like that INI config either, but was worried about introducing yet another breaking change in this PR. Are there other config files that y'all are using for other scimma projects that we can standardize on? That may be hard to do for other reasons, but wanted to bring that up in case there was some flexibility there. TOML vs YAML?

hop/auth.py
myNameIsPatrick: Sounds good to me. This may be applicable to both comments but I can default to `SCRAM-SHA-512` if nothing is set.

hop/auth.py
myNameIsPatrick: Sounds good to me.

tests/test_io.py
myNameIsPatrick: That's a good idea, I'll go add that in when I get a chance.

tests/test_io.py
myNameIsPatrick: > What does this test actually test? I think all that's really getting tested is that stream.open does not raise, and that it returns the contents of mock_consumer.stream (in other words, that your mocking works). One issue with using mocks is that you can wind up testing that your mocking library is working, but not whether your code is correct.

> Maybe that's sufficient here, but it's a little fishy to me - it doesn't test that start_at or persist or the URL are handled correctly, just that these options don't raise an exception.

Yeah, that's right. All this test does is check that `stream.open` doesn't raise. It's better than nothing, but far from satisfactory. It could be improved by checking that certain calls were made to `_Consumer.stream` like you mentioned, but still doesn't feel sufficient. The biggest issue that I've encountered here is that because `_Consumer` subclasses `adc.Consumer` and extends the behavior of `stream()`, I'm not able to mock the parent's class' `stream()`. Maybe not at all, maybe not without considerable work. That would be better because then I can mock sufficiently upstream that I could test the right behavior sufficiently assuming that anything living in `adc` is correct. I kinda went down the refactoring road a bit to split up all the work into these pack/unpack static methods that I could test in isolation, but didn't go any further than that.

I guess the only two things this actually tests is if there's a bug introduced that egregious enough that it raises or the API has changed enough where it would start to fail. Beyond that...nah.

Out of the improvements you suggested, I think a combination of adding integration tests and letting the CLI tests handle more of the heavy lifting sound the most reasonable to me. Adding in integration tests is probably out of scope for this PR but would be good to raise an issue for and tackle soon. Does that sound reasonable?

myNameIsPatrick:
Thanks @spenczar for the thorough review! I'll start incorporating some of those changes you requested.

doc/user/commands.rst
```diff
@@ -8,14 +8,29 @@ Commands
...
 :code:`hop publish`
 ~~~~~~~~~~~~~~~~~~~~~~
 
-This command allows a user to parse an `RFC 822 formatted GCN circular <https://gcn.gsfc.nasa.gov/gcn3_circulars.html>`_, an XML formatted `GCN/VOEvent notice <https://gcn.gsfc.nasa.gov/tech_describe.html>`_, or an unstructured string or text file and publish it as via Kafka. Structured messages
-such as GCN circulars and VOEvents are published as JSON-formatted text.
+This command allows a user to parse various structured and unstructured messages, including:
+
+* `RFC 822 formatted GCN circular <https://gcn.gsfc.nasa.gov/gcn3_circulars.html>`_
+* An XML formatted `GCN/VOEvent notice <https://gcn.gsfc.nasa.gov/tech_describe.html>`_
+* An unstructured string or text file and publish it as via Kafka.
```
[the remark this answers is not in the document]
myNameIsPatrick: Will fix this shortly.

setup.py
```diff
@@ -8,7 +8,7 @@
 
 # requirements
 install_requires = [
-    "adc-streaming == 0.3.1",
+    "adc-streaming >= 1.0.0",
```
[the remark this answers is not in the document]
myNameIsPatrick: Good call, I always forget that we publish our own conda package.

tests/test_io.py
```diff
@@ -4,29 +4,141 @@
...
+    content = {}
+    for field in fields(message_model):
+        content.update({field.name: "test"})
+    return content
 
-        # verify defaults are stored correctly
-        assert stream._options["format"] == format
-        assert stream._options["start_at"] == start_at
-        assert stream._options["config"] == config
+# test the subscribe classifier for each message format
```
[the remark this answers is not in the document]
myNameIsPatrick: Nice catch, I'll fix that up.

hop/io.py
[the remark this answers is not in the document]
spenczar: Okay, I understand.

I don't think that's the API I would design - you save on typing `auto=load_auth(), persist=True` twice, but there isn't much other benefit - but it makes sense. As long as you're thinking about it, I'd say you should use your judgement.

hop/io.py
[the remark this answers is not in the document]
spenczar: I think a logged warning is definitely a good approach, since it will lead users to actually report the issue. I was thinking just some sort of `blob.missing_schema = True` would be fine for identifying it programmatically.

hop/io.py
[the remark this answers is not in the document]
spenczar: Oh, pluggy looks really nice. That would make it so you can just `pip install <some-serialization-plugin>`, which is really simple. I like it.

hop/auth.py
spenczar: Up to you, I think it looks kind of messy and doesn't age as well as putting things in a separate file. It's definitely an aesthetic thing, though.

hop/auth.py
spenczar: Right, changing it in a separate PR seems better.

YAML or TOML - I don't have a strong preference, but lean slightly towards TOML. YAML can have very suprising behavior because the spec is so ridiculously enormous. For example, there's the classic:
```yaml
python: 3.5.3
postgres: 9.3
```
which evaluates to 
```python
{
    "python": "3.5.3",
    "postgres": 9.3000000001
}
```

See https://www.arp242.net/yaml-config.html for more. On the other hand, TOML is kind of newish and foreign to many people, and it has its own weirdnesses. Nested dictionaries are _really_ weirdly expressed in TOML, in particular.

So there's no real great options. When projects really want to take this seriously, I think they end up inventing their own configuration language for that reason - that's why [nginx](https://nginx.org/en/docs/beginners_guide.html#conf_structure) and [Apache](https://httpd.apache.org/docs/2.4/configuring.html) and [Varnish](https://varnish-cache.org/docs/trunk/users-guide/vcl.html) etc all have their own configuration languages, complete with their own parsers.

Of course, we shouldn't do that, I just think it's kind of an interesting aside :) I think TOML or YAML are both fine.

tests/test_io.py
spenczar: Yep, I think that sounds right to me too :+1: .

hop/io.py
```diff
@@ -3,44 +3,218 @@
...
-        opts = {**self._options, **kwargs}
-        return streaming.open(*args, **opts)
+        try:
+            format = message["format"].upper()
+            content = message["content"]
+        except (KeyError, TypeError):
+            raise ValueError("Message is incorrectly formatted")
+        else:
+            if format in cls.__members__:
+                return cls[format.upper()].value(**content)
```
spenczar, line 133: I just noticed this `**content` bit. I think that means that we're expecting `content` to always be a dictionary. What if it's arbitrary bytes?

For example, what if it's something like base64-encoded Avro?

```
{
    "format": "LSST_ALERT_AVRO",
    "content": "AAAAAADQ3A2St+vd2gHAsIkBAAAAZCPYoNLS7EAGdGVkAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAYHeXSQAA+EIAAPhCABxCRgAcQk",
}
```

Maybe we're saying that `content` **must** be a JSON dictionary? That works, but we should be clear about it I think.

hop/auth.py
myNameIsPatrick: I axed it and will be removing the rest in the future PR to not overwhelm this one.

hop/auth.py
myNameIsPatrick: I ended up setting the default (`SCRAM-SHA-512`) in the previous comment but will tackle all the defaults/configuration stuff in a future PR.

doc/user/commands.rst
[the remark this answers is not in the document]
myNameIsPatrick: I ended up rewriting that bullet point so it matches better with the the rest.

hop/io.py
myNameIsPatrick: That is definitely a bug. All the unstructured messages I was testing with were JSON dictionaries which may explain why I didn't catch it. I definitely want to support any arbitrary bytes, but would have to think about how to make that work here, since that would fail out in the current implementation.

hop/auth.py
myNameIsPatrick: What's wrong with a little remote code execution? Seriously though, I have no strong preference either. I've used YAML much more but I've also been bitten by that example you posted above. Let's try to plan out the configuration sometime after this is in.

hop/io.py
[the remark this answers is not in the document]
myNameIsPatrick: Awesome. We can start planning this out in the near future for users to extend our current classes in a nice way.

hop/io.py
myNameIsPatrick: So, I have something working that should support arbitrary content as long as it's JSON serializable. That doesn't include bytes, but does include base64 encoded strings. I think if we want to support raw bytes we'd have to auto-detect it and decode/encode as needed or move to Avro which I believe has a native bytes type.

hop/io.py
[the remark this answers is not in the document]
myNameIsPatrick: Added this in just now, along with supporting arbitrary JSON-serializable unstructured messages. An extra conditional was added to handle the unstructured case differently from the rest (the previous implementation only supported dictionaries unintentionally). In the case the message format is not recognized, a `MessageBlob` is returned with `missing_schema = True`.

tests/test_io.py
myNameIsPatrick: This has now been added.

tests/test_io.py
```diff
@@ -150,3 +151,43 @@ def test_pack(circular_msg, circular_text):
...
+    {"format": "blob", "content": content_mock(MessageBlob)},
+])
+def test_pack_unpack_roundtrip(message, message_parameters_dict, caplog):
+    format = message["format"]
+    content = message["content"]
+
+    # load test data
+    shared_datadir = Path("tests/data")
+    test_filename = message_parameters_dict[format]["test_file"]
+    test_file = shared_datadir / "test_data" / test_filename
```
spenczar, line 168: whoa, TIL about pathlib

hop/io.py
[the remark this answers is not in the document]
spenczar: Made https://github.com/scimma/hop-client/issues/84 to track

hop/auth.py
spenczar: Created https://github.com/scimma/hop-client/issues/85 to track

hop/io.py
spenczar: Right, if it's truly raw bytes, I don't think it we'd have it anyway: we wouldn't be able to `json.loads` the message payload, so we wouldn't have the `message["content"]` anyway. I think its fair to require base64-encoding as long as we have a JSON wrapper.

hop/publish.py
```diff
@@ -4,64 +4,38 @@
...
-
     parser.add_argument(
         "-f",
         "--format",
-        type=str,
-        default="blob",
-        help="Specifies the format of the message(s), such as 'circular' or 'voevent'. "
-        "Specify 'blob' if sending an unstructured message. Default: 'blob'.",
+        choices=io.Deserializer.__members__,
+        default=str(io.Deserializer.BLOB),
```
bfc5288, line 25: This needs to be `default=str(io.Deserializer.BLOB.name),`, otherwise the default format isn't loaded correctly:
```
#: hop publish kafka://scimma-server:9092/test tests/data/example.gcn3   # don't specify a format, so it should be a BLOB

Traceback (most recent call last):
  File "/opt/conda/bin/hop", line 10, in <module>
    sys.exit(main())
  File "/opt/conda/lib/python3.7/site-packages/hop/__main__.py", line 72, in main
    args.func(args)
  File "/opt/conda/lib/python3.7/site-packages/hop/publish.py", line 35, in _main
    loader = io.Deserializer[args.format]
  File "/opt/conda/lib/python3.7/enum.py", line 352, in __getitem__
    return cls._member_map_[name]
KeyError: 'Deserializer.BLOB'
```

hop/models.py
```diff
@@ -148,41 +148,41 @@ def from_email(cls, email_input):
...
 
         Returns:
-            dict: the dict representation of the message
+            The dictionary representation of the message.
 
         """
         return asdict(self)
 
-    def wrap_message(self):
+    def serialize(self):
```
bfc5288, line 185: I believe that `wrapped_message` should have `content` be assigned to `self.content`:
```
wrapped_message = {"format": "blob", "content": self.content}
```

Otherwise, `subscribe.print_message` fails when handling `BLOB` messages:
```
# hop publish kafka://scimma-server:9092/test-blob-file02 -f BLOB tests/data/example.gcn3
# hop subscribe kafka://scimma-server:9092/test-blob-file02 -s EARLIEST -p
Traceback (most recent call last):
  File "/opt/conda/bin/hop", line 10, in <module>
    sys.exit(main())
  File "/opt/conda/lib/python3.7/site-packages/hop/__main__.py", line 72, in main
    args.func(args)
  File "/opt/conda/lib/python3.7/site-packages/hop/subscribe.py", line 69, in _main
    print_message(message, args.json)
  File "/opt/conda/lib/python3.7/site-packages/hop/subscribe.py", line 29, in print_message
    print(str(message_model))
TypeError: __str__ returned non-string (type dict)
```

hop/publish.py
```diff
@@ -4,64 +4,38 @@
...
-            loader = model_loader[args.format]
-        else:
-            warnings.warn(
-                "Warning: format not recognized. Sending as unstructured blob")
-            loader = model_loader["blob"]
-
         for message_file in args.message:
-            message_model = loader(message_file)
-            s.write(message_model.wrap_message())
+            message = loader.load_file(message_file)
```
bfc5288, line 40: This forces the published message to be a file -- is this desired? This precludes us from sending string messages from CLI:
```
hop publish kafka://scimma-server:9092/test-blob-string01 -f BLOB "plain text string"
Traceback (most recent call last):
  File "/opt/conda/bin/hop", line 10, in <module>
    sys.exit(main())
  File "/opt/conda/lib/python3.7/site-packages/hop/__main__.py", line 72, in main
    args.func(args)
  File "/opt/conda/lib/python3.7/site-packages/hop/publish.py", line 40, in _main
    message = loader.load_file(message_file)
  File "/opt/conda/lib/python3.7/site-packages/hop/io.py", line 145, in load_file
    return self.value.load_file(input_file)
  File "/opt/conda/lib/python3.7/site-packages/hop/models.py", line 226, in load_file
    with open(filename, "r") as f:
FileNotFoundError: [Errno 2] No such file or directory: 'plain text string'
```

We may not want this anyways, but checking just in case (especially since `models` has the tooling to read strings via `load()`).

hop/models.py
```diff
@@ -148,41 +148,41 @@ def from_email(cls, email_input):
...
 class MessageBlob(object):
     """Defines an unformatted message structure.
 
-    This is included as a dataclass to mirror the implementation of structured formats.
+    This is included to mirror the implementation of structured formats.
 
     """
 
     content: str
+    missing_schema: bool = False
```
bfc5288, line 174: FYI -- this shows up as a key/value whenever doing `subscribe ... --json` for `CIRCULAR` and `BLOB` formats, but not `VOEVENT`. Maybe we could ignore it if it's false, but this is a small detail for now.

hop/publish.py
myNameIsPatrick: Thanks, this should be fixed now.

hop/models.py
myNameIsPatrick: Nice catch, I missed this in testing and it appears to be a blind spot in our CLI tests.

hop/models.py
myNameIsPatrick: I went ahead with what you suggested, as in not show this attribute unless it's set to true. Running some tests with this, this only popped up in the `BLOB` format but not the `CIRCULAR` format. As in, if you publish a circular as a blob, it'll show up, but publishing it as a circular does not.

hop/publish.py
myNameIsPatrick: That's a good point, I hadn't appreciated that's how you wired it up when you added the blob functionality. I think being able to pass in strings is good to have, but I'm a bit wary of having this functionality present with one format but not another without it being more clear how it's being handled. As in, I think it may be clearer if `load()` has consistent behavior across all messages.

What may be nice is to be able to pipe input into `hop publish` in the same way other UNIX commands are done and this may tie nicely into passing in strings, as in:
`cat stringlist.txt | hop publish kafka://host:port/topic`

This would in principle work with all formats, as long as each string token is a message to be published. I think what we'd need to do is to modify the message argparse option to possibly take in zero arguments as well and check if `sys.stdin.isatty()` to read terminal input from stdin. What do you think?

hop/models.py
bfc5288: That looks about right. Though with `subscribe`, I now don't see any reference to `missing_schema` (even for messages published with no specified format).

At first pass, I think this is because I can't trigger the [missing_schema line](https://github.com/scimma/hop-client/blob/update/adc_api/hop/io.py#L138) with either `subscribe` or `stream`, as any message that is lacking a specified format is "rescued" into a `BLOB` format, though I might be missing something.

(Though in general, I don't think this is necessarily bad behavior)

hop/publish.py
bfc5288: I agree, and I think implementing the UNIX-like piping method would be nice. I think it is out-of-scope for this PR, so I made #86.

hop/models.py
myNameIsPatrick: Out of curiosity, how are you producing the poorly-formatted messages? The way it's set up, nothing you can send through hop will trigger missing_schema=True or raise the incorrectly formatted error.

The missing_schema line can be triggered by (1) extending the Deserializer to take in extra formats that a downstream subscriber doesn't have access to, or (2) sending in a message with the form {"format": "something", "content": "something"} which is JSON serialized and utf-8 encoded. Example:

```
[patrick@localhost ~/workspace/projects/github/scimma/hop-client] [update/adc_api] $ python
Python 3.7.4 (default, Jul  9 2019, 16:32:37) 
[GCC 9.1.1 20190503 (Red Hat 9.1.1-1)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import confluent_kafka, json
>>> p = confluent_kafka.Producer({"bootstrap.servers": "localhost:9092", "group.id": "hey"})
>>> msg = json.dumps({"format": "bad", "content":{"hey": "you"}}).encode("utf-8")
>>> p.produce("test3", msg)
>>> p.flush(0)
0
>>> 
[patrick@localhost ~/workspace/projects/github/scimma/hop-client] [update/adc_api] $ hop subscribe kafka://localhost:9092/test3 -s EARLIEST --no-auth --json
Message format BAD not recognized, returning a MessageBlob
{"content": {"hey": "you"}, "missing_schema": true}
```

The incorrectly-formatted error can be triggered in a similar way, except that the message format is something that the Consumer is not able to unpack.

bfc5288: Got it. I wasn't extending or sending a JSON-formatted message with a bad format, just publishing a message that would be interpreted as `BLOB`. That all looks good to me now.

bfc5288:
This looks ready to me. Excellent work putting everything together! Looking forward to doing a release of all these changes soon.

myNameIsPatrick:
Thanks @bfc5288 and @spenczar for the very thorough review and good discussions that came out of this, including new user stories (#84, #85 and #86)!

Issue #85, opened 2020-07-02
UX: Design and implement a configuration file format


spenczar:
For background and motivation, see https://github.com/scimma/hop-client/pull/82#discussion_r447194917.

### Description
As a user, I want to configure the hop client without needing magic invocations like "sasl.mechanism=SCRAM-SHA-512".

I want a configuration language which supports comments.

I want a configuration language which is easy readable by humans.

I want a configuration language which is easily diffed in pull requests.

I want a configuration language which supports validation without execution.

I want to configure my username and password in one place for hop-client.

I want to be able to upgrade from an old configuration file to a new one easily.

### Definition of Done
Support a new configuration file format for publishing and subscribing with hop.

myNameIsPatrick:
Based on the spec laid out by @spenczar in https://github.com/scimma/hop-client/pull/82#discussion_r447194917, I am proposing the following minimal configuration format using TOML:

```
[auth]
username = "username"
password = "..."
```

I'm a bit hesitant whether we should expose the following properties as well, but it may provide some flexibility while doing development and while all the security details are not set in stone yet (although please correct me if I'm wrong on this):

```
[auth]
...
protocol = "SASL_SSL"
mechanism = "SCRAM-SHA-512"
ssl_ca_location = "/path/to/cacert"
```

I went along with TOML rather than YAML, partly due to the smaller spec, partly because it's likely the configuration will be mostly flat and may look nicer, and partly due to python's adoption of TOML for project management (see https://www.python.org/dev/peps/pep-0518/).

Any comments/suggestions are welcomed.

dpetravick:
can the config file support multiple usernames?

myNameIsPatrick:
@dpetravick, the config format certainly can but the current spec doesn't allow for that. It would be nice to support, however. I'd have to think of a clean way of adding that in that looks sensible.

spenczar:
What would multiple usernames mean? I don't see how you could connect to the Kafka broker as multiple users at the same time.

myNameIsPatrick:
I had interpreted that to be supporting multiple user profiles that you can switch between at runtime (with some default).

spenczar:
I think we should default to you-aint-gonna-need-it; that adds a lot of complexity to the config format and to the code, and I don't see the benefit. I could be missing it, though.

shereenElSayed:
[...]

PR #89, opened 2020-07-15
Switch auth configuration to TOML

myNameIsPatrick:
## Description

This PR switches the configuration format from librdkafka-style conf to TOML, based on #85. In addition, the `Auth` class in `auth.py` has been modified to expose `SCRAM-SHA-512` as the default SASL mechanism. Finally, upon setup to generate a new configuration file, the user is prompted for a username and password (using `getpass.getpass()` to hide the password).

## Checklist

### DIFF
--- a/hop/auth.py
+++ b/hop/auth.py
@@ -8,23 +8,43 @@ import errno
 import logging
 import os
 
+import toml
+
 from adc import auth
 
 DEFAULT_AUTH_CONFIG = """\
-security.protocol=SASL_SSL
-sasl.mechanism=SCRAM-SHA-512
-sasl.username=username
-sasl.password=password
+[auth]
+username = "username"
+password = "password"
 """
 
 logger = logging.getLogger("hop")
 
-# expose auth options from adc
-SASLAuth = auth.SASLAuth
 SASLMethod = auth.SASLMethod
 
-# map default auth option to Auth
-Auth = auth.SASLAuth
+
+class Auth(auth.SASLAuth):
+    """Attach SASL-based authentication to a client.
+
+    Returns client-based auth options when called.
+
+    Parameters
+    ----------
+    user : `str`
+        Username to authenticate with.
+    password : `str`
+        Password to authenticate with.
+    ssl : `bool`, optional
+        Whether to enable SSL (enabled by default).
+    method : `SASLMethod`, optional
+        The SASL method to authenticate, default = SASLMethod.SCRAM_SHA_512.
+        See valid SASL methods in SASLMethod.
+    ssl_ca_location : `str`, optional
+        If using SSL via a self-signed cert, a path/location
+        to the certificate.
+    """
+    def __init__(self, user, password, ssl=True, method=SASLMethod.SCRAM_SHA_512, **kwargs):
+        super().__init__(user, password, ssl=ssl, method=method, **kwargs)
 
 
 def get_auth_path():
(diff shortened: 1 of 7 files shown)


hop/auth.py
```diff
@@ -115,6 +142,10 @@ def _main(args):
         if os.path.exists(authfile) and not args.force:
             logger.warning("configuration already exists, overwrite file with --force")
         else:
+            logger.info("generating configuration with user-specified username + password")
             os.makedirs(os.path.dirname(authfile), exist_ok=True)
+
+            user = input("Username: ")
             with open(authfile, "w") as f:
-                f.write(DEFAULT_AUTH_CONFIG)
+                f.write(DEFAULT_AUTH_CONFIG.format(username=user, password=getpass.getpass()))
```
spenczar, line 150: Kinda feels like this should be a `toml.dump` call... I guess the advantage of this is that you get a `DEFAULT_AUTH_CONFIG` value which helps to document things?

hop/auth.py
```diff
@@ -5,26 +5,48 @@
...
+        Whether to enable SSL (enabled by default).
+    method : `SASLMethod`, optional
+        The SASL method to authenticate, default = SASLMethod.SCRAM_SHA_512.
+        See valid SASL methods in SASLMethod.
+    ssl_ca_location : `str`, optional
+        If using SSL via a self-signed cert, a path/location
+        to the certificate.
+    """
+
+    def __init__(self, user, password, ssl=True, method=SASLMethod.SCRAM_SHA_512, **kwargs):
```
spenczar, line 48: Why `**kwargs` instead of explicitly including `ssl_ca_location`?

hop/auth.py
myNameIsPatrick: You're right, this is left over from the previous setup writing .ini files to disk. I can switch this to pass a dict to `toml.dump`.

hop/auth.py
myNameIsPatrick: No good reason, just didn't want to unintentionally miss any options.

hop/auth.py
```diff
@@ -5,26 +5,48 @@
...
 logger = logging.getLogger("hop")
 
-# expose auth options from adc
-SASLAuth = auth.SASLAuth
 SASLMethod = auth.SASLMethod
 
-# map default auth option to Auth
-Auth = auth.SASLAuth
+
+class Auth(auth.SASLAuth):
```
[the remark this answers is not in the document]
myNameIsPatrick: Not much, the only intended purpose is to set the default method to be `SCRAM-SHA-512` instead of the default in `adc-streaming`. This could be the default in `adc-streaming` or maybe there's an easier way to do this if we wanted to decouple them in any way.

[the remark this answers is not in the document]
spenczar: Ah, yeah, I see. What you have here makes some sense. You could do it as a function too (`def default_auth` or something), but I'm not sure that's much better.

Could you add a little comment explaining that this class exists to set our desired defaults? I have been thinking a lot lately about https://brooker.co.za/blog/2020/06/23/code.html and it's inspired me to push for comments which describe *design decisions*, rather than just comments that describe what code does, and this is a good example.

hop/auth.py
myNameIsPatrick: Fixed in the latest commit.

hop/auth.py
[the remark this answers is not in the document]
myNameIsPatrick: Thanks @spenczar, that was a good read and a nice approach to take for documenting code. I added some reasoning above the class definition. Let me know if that sounds good or whether you'd prefer it somewhere else (docstring or otherwise).

myNameIsPatrick:
@spenczar, just wanted to check in to see if you wanted to see any changes or if everything looks good here.

myNameIsPatrick:
@spenczar, thanks for the quick response and for the review! I'll go ahead and merge this in.

Issue #86, opened 2020-07-03
Implement UNIX-like piping to publish messages 


bfc5288:
### Description

As a user, I would like to pipe raw strings directly from the CLI or from a file into `hop publish` in a UNIX-like fashion, e.g.:
```
"This is my message" | hop publish kafka://host:port/topic
```
or
```
cat stringlist.txt | hop publish kafka://host:port/topic
```

See [discussion](https://github.com/scimma/hop-client/pull/82#discussion_r449690982) in #82. 

### Definition of Done
A command like `cat stringlist.txt | hop publish kafka://host:port/topic` publishes the strings from `stringlist.txt` as message(s), while also respecting the formatting of the strings (if any) and mapping this to the corresponding message datamodel.

### Tasks/Items:
- [x] Change CLI argument settings as needed
- [ ] Verify for each format datamodel
- [ ] Write unit tests

PR #90, opened 2020-07-17
Piping input into hop publish

myNameIsPatrick:
## Description

This PR implements https://github.com/scimma/hop-client/issues/86, to pipe messages into `hop publish`. This allows things like:

```
cat stringlist.txt | hop publish kafka://host:port/topic
```
and also
```
hop publish kafka://host:port/topic < stringlist.txt
```

A few gotchas about this is that you need to be careful to not add keyword arguments between positional arguments, since argparse will get confused and try to process one of those keyword arguments as a message.

Another is this likely won't make sense to use for circulars due to the format of said messages. We should probably add a sanity check that disables this message format when trying to parse messages from stdin.

Finally, I don't have a good way of adding in test coverage for this, since emulating terminal input seems to be much more difficult to do than mocking a file read.

## Checklist

* [ ] Add/update pytest-style tests in `/tests`, ensuring sufficient code coverage.

### DIFF
--- a/hop/publish.py
+++ b/hop/publish.py
@@ -31,6 +31,13 @@ def _main(args):
             message = loader.load_file(message_file)
             s.write(message)
 
+        # check if stdin is connected to TTY device
         if not sys.stdin.isatty():
-            for message in sys.stdin.read().splitlines():
-                s.write(loader.load(message))
+            messages = sys.stdin.read().splitlines()
+
+            if messages:
+                assert args.format == io.Deserializer.BLOB.name, \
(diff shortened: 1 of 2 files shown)


myNameIsPatrick:
@bfc5288, this should be ready to look at. I added an explicit check that this can only be used for the BLOB format, since it's too easy to have this do the wrong thing for circulars or VOEvents. This may be revisited when we add message types where something like this won't cause user-facing issues.

myNameIsPatrick:
@bfc5288, definitely, for sure. Thanks for the review!

PR #92, opened 2020-07-23
Consistency between CLI and python API

myNameIsPatrick:
## Description

This PR provides a more consistent experience between the CLI and python API. For example, for the CLI, auth credentials were automatically loaded in from a configuration file, but one needs to explicitly pass in an `Auth` instance for the python API. This change in behavior for the two interfaces has been removed. Now, the same defaults are present in both, and a `Stream()` instance by default uses authentication.

@spenczar, while I was doing this, I was thinking about the conversation from a previous PR (https://github.com/scimma/hop-client/pull/82#discussion_r447181113) on why options can be passed in both the `Stream` constructor and for `Stream.open()`. I'm rethinking my original position and leaning towards "ya ain't gonna need it" and instead just giving the default `Stream` instance sensible defaults. So that's all gone now, we can only pass in extra configuration upon class instantiation.

There was a few extra hoops to jump through in terms of making the default behavior to provide authentication in the python API, partly to allow boolean flags or an `Auth` instance to turn on/off authentication. The other part is due to this convenience class `stream` in hop's `__init__.py`, which would try to auto-load authentication upon importing hop. I opted for an approach that lazily loads authentication upon the first call to `Stream.open()` and caches it for further reuse. It also raises an error if it couldn't load credentials to authenticate. It's a...little extra complexity but I like being able to do something like this if I don't want to change any defaults:

```python
from hop import stream

with stream.open("kafka://localhost:9092/topic", "w") as s:
    stream.write(message)
```

Finally, I updated the docs accordingly for the change in behavior for the python API.

## Checklist

### DIFF
--- a/hop/io.py
+++ b/hop/io.py
@@ -1,5 +1,6 @@
 from contextlib import contextmanager
 from dataclasses import dataclass
+from functools import lru_cache
 import getpass
 from enum import Enum
 import json
@@ -11,7 +12,7 @@ import warnings
 
 from adc import consumer, errors, kafka, producer
 
-from .auth import load_auth
+from .auth import get_auth_path, load_auth
 from . import models
 
 logger = logging.getLogger("hop")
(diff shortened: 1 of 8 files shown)


hop/io.py
```diff
@@ -25,30 +27,51 @@ class Stream(object):
...
-    def __init__(self, auth=None, start_at=None, persist=None):
-        self.auth = auth
+    def __init__(self, auth=True, start_at=StartPosition.LATEST, persist=False):
+        self._auth = auth
         self.start_at = start_at
         self.persist = persist
 
-    def open(self, url, mode="r", auth=None, start_at=None, persist=None, metadata=False):
+    @property
+    @lru_cache(maxsize=1)
```
spenczar, line 44: Its a bit of a surprise to me that this is the decorator of choice; why not `functools.cached_property`?

myNameIsPatrick: That's awesome, I had no idea that existed! It looks like it's new as of python 3.8+ so it may be worth holding off on this until we officially drop 3.6 and 3.7.

myNameIsPatrick:
Thanks @spenczar for the review and showing me something new! I'll go ahead and merge this.

Issue #84, opened 2020-07-02
UX: Support message format plugins


spenczar:
### Description
As a developer, I want to add my own message formats through plugins.

We have message formats, and we support a small set of them. But letting users add more through pip-installable plugins would be really great.

See discussion in https://github.com/scimma/hop-client/pull/82/files#r447343536; pluggy might be a good approach here.

### Definition of Done
Have an external package which provides a demonstration plugin, adding support for some new message format. It can be a test format, or it could be something real, like LSST alerts.

myNameIsPatrick:
Was playing around with this for another project and have convinced myself that using pluggy would be definitely doable and would work nicely with the client.

PR #96, opened 2020-08-04
Allow custom message formats via plugins

myNameIsPatrick:
## Description

This PR implements #84, to support external message plugins via pluggy. In doing so, I had created an abstract message type, `MessageModel` which lives in `models.py` which other message types can subclass to implement the `load()` functionality and override any methods as needed. I also renamed the `MessageBlob` as `Blob` to re-use the base implementation to generate the message wrapper.

There's also a new module, `plugins.py` which contains hooks for implementing various message plugins. Various plugins are registered in `models.py` and can also be registered in an external library as long as it's registered in the right way. The internal messages are registered via:

```
from hop import plugins

...

@plugins.register 
def get_models(): 
    return { 
        "voevent": VOEvent, 
        "circular": GCNCircular, 
        "blob": Blob, 
    } 
```

Then they get automatically discovered in the `Deserializer` in `hop.io` as long as it registers plugins as key-value pairs in a function called `get_models()` with this decorator. The other magic happens in the `Deserializer`, which has been rewritten using the functional Enum API to allow this. It uses a helper function to load various plugins and a mixin class to add in the extra methods to the Enum class.

One thing that is missing currently is documentation on how to register an external plugin, which I'll add in but I wanted to open this up now since the rest of the PR is ready. As a summary on how to register an external plugin, essentially you need to do this:

* Add an entrypoint to `setup.py` called `hop_plugin` which points to the module where the implementations are.
* Implement a message model which subclasses `models.MessageModel` and defines at least the `load()` functionality.
* Register the plugin via the method above.

## Checklist


hop/io.py
```diff
@@ -111,11 +114,32 @@ def open(self, url, mode="r", metadata=False):
...
+    # set up plugin manager
+    manager = pluggy.PluginManager("hop")
+    manager.add_hookspecs(plugins)
+
+    # load in base models
+    manager.register(models)
+
+    # load external models
+    try:
+        manager.load_setuptools_entrypoints("hop_plugin")
```
spenczar, line 130: Note for later - make sure to document how to define conda recipes that provide entry_points. I thought that it wouldn't be possible, but it turns out you can do it: https://stackoverflow.com/questions/50106303/how-to-i-create-an-entry-point-to-a-conda-package-application

hop/io.py
```diff
@@ -111,11 +114,32 @@ def open(self, url, mode="r", metadata=False):
...
+    except Exception:
+        logger.warning("WARNING: could not load external hop models, verify installed plugins")
```
spenczar, line 132: Please remove the `WARNING: ` prefix. The logger should say WARN for you anyway.

I think this message will be pretty cryptic for most users. What does "verify installed plugins" mean? Picture a stumped user getting that message; I don't think they'd know what to do.

hop/io.py
```diff
@@ -111,11 +114,32 @@ def open(self, url, mode="r", metadata=False):
...
 
+    # add all registered plugins to registry
+    registered = {}
+    for model_plugins in manager.hook.get_models():
+        registered.update({name.upper(): model for name, model in model_plugins.items()})
+
```
spenczar, line 138: Personally, I find this a bit more clear (but just a bit):
```python
registered = {}
for model_plugins in manager.hook.get_models():
    for name, model in model_plugins.items():
        registered[name.upper()] = model
```

---

I think we should warn when there are name collisions, not silently pick the last one. That'd be a nasty thing to try to debug.

hop/io.py
myNameIsPatrick: Those are good points, I'll think about a better way of wording that so that a user will know what's going on.

hop/io.py
myNameIsPatrick: Will do, I'll update that code snippet with your suggestion. Regarding the name collision, that's a good idea. I'll add a check to see if the key exists while looping over the various plugins and warn if there are duplicates.

hop/io.py
myNameIsPatrick: Will do. That's something I didn't quite get from defining an entry point via conda. If you set up entry points via setuptools in `setup.py`, `setup.cfg` or `pyproject.toml`, does that propagate in the right way through a conda recipe or do you need to also specify it explicitly?

myNameIsPatrick:
Thanks @spenczar for the review and the nice feedback. 

> I think it would be a good idea to try writing a plugin in a separate package to test everything out end-to-end, including pushing it to PyPI.

I gave that a whirl locally, minus publishing the actual package to PyPI. It's a barebones library which registers a custom message format via the `hop_plugin` entry point and will be demoing that at the next sprint review.

hop/io.py
myNameIsPatrick: As an update, it seems like the answer to that question is yes. I'm looking at various feedstocks for pytest plugins for conda which use the same discovery mechanism and it looks like they don't define the entry points within the recipe in order for this to work out: https://github.com/conda-forge/pytest-console-scripts-feedstock/blob/master/recipe/meta.yaml

PR #101, opened 2020-08-13
Modify formatting for CLI when calling --help

myNameIsPatrick:
## Description

This PR modifies the formatting for various help commands in the CLI to generally look nicer. There's also a few auxiliary items I have modified which I'll enumerate below. First, the CLI output:

#### Before:

```
 $ hop --help
usage: hop [-h] [--version] <command> ...

optional arguments:
  -h, --help  show this help message and exit
  --version   show program's version number and exit

Commands:
  <command>
    configure
              configuration utilities
    publish   parse and publish messages
    subscribe
              receive and parse messages
    version   list all the dependencies' versions
```

```
 $ hop configure --help
usage: hop configure [-h] <command> ...

Configuration utilities.

optional arguments:
  -h, --help  show this help message and exit

Commands:
  <command>
    locate    display configuration path
    setup     set up configuration
```

#### After:

```
 $ hop --help
usage: hop [-h] [--version] <command> ...

optional arguments:
  -h, --help          show this help message and exit
  --version           show program's version number and exit

commands:
    configure         configuration utilities
    publish           publish messages
    subscribe         subscribe to messages
    version           list all the dependencies' versions
```

```
 $ hop configure --help
usage: hop configure [-h] <command> ...

Configuration utilities.

optional arguments:
  -h, --help          show this help message and exit

commands:
    locate            display configuration path
    setup             set up configuration
```

The two main changes is this removal of the `<command>` line between the commands and the header and the spacing between commands and the description. This is handled by an internal function in `hop/__main__.py`, `_SubcommandHelpFormatter` which modifies the format of the help when passed into `formatter_class` in an argparse-based parser instance. Going a bit into the technical side, argparse seems to have a strange interaction between the metavar kwarg and subparsers which I'm essentially reverting so the spacing looks the same as if you didn't specify this.

The other auxiliary changes I have made are in `configure.py`. A few changes in the docstrings were made to match the rest of the codebase, and I added the `required` flag in the configure subparser so that you can't call `hop configure` by itself. This removes the need to check explicitly in the main function, and will warn the user accordingly of the proper usage.

## Checklist


hop/__main__.py
```diff
@@ -8,22 +8,51 @@
 from . import version
 
 
-def append_subparser(subparser, cmd, func):
+def append_subparser(subparser, cmd, func, formatter=False):
```
spenczar, line 11: A boolean for `formatter` seems slightly odd. I would have expected it to be a formatter class, so you pass in `formatter=_SubcommandHelpFormatter`.

hop/__main__.py
```diff
@@ -8,22 +8,51 @@
...
+        )
+    else:
+        parser = subparser.add_parser(cmd, help=help_, description=desc)
 
     parser.set_defaults(func=func)
+
     return parser
 
 
+class _SubcommandHelpFormatter(argparse.RawDescriptionHelpFormatter):
```
shereenElSayed, line 40: What do you think if this class is moved to cli.py? I think it fits more there

hop/__main__.py
shereenElSayed, line 11: Maybe we can change formatter parameter to apply_formate?

myNameIsPatrick:
@spenczar, @shereenElSayed, I see two different approaches on how to improve how `formatter` is handled within `append_subparser()`:

1. Pass in a formatter class instead of a boolean, via `formatter` or possibly `formatter_class` to match the kwarg being passed to the parser.
2. Keep `formatter` as a boolean but change the name to `apply_format`.

I would be slightly in favor of the first as it's a bit more flexible but otherwise don't have any strong opinions. Thoughts?

hop/__main__.py
myNameIsPatrick: I agree with that. I found a technical issue with doing that however, which is kinda annoying. The main issue is that adding it to the `cli` module adds in a circular dependency as `__main__.py` imports from all modules defining a `_main()` function while the other modules import from `cli.py` to grab commonly used CLI utilities.

shereenElSayed: hmm I see, what do you think of a utility module?

shereenElSayed:
@myNameIsPatrick I'd go with the first option

hop/__main__.py
myNameIsPatrick: That's a good idea, I added this in the last batch of commits.

myNameIsPatrick: I modified this in the latest batch of commits.

hop/__main__.py
myNameIsPatrick: I like it. I added this class and the subparser function to `hop.utils.cli` in the latest batch of commits.

PR #102, opened 2020-08-19
Expose manual committing of messages + allow use of stream.open() without context manager

myNameIsPatrick:
## Description

Basically, the big thing here is a rework of the internal Producer/Consumer classes used within `stream.open()`.
    
These features can be split into allowing manual committing of messages, changing the API slightly for clarity, and making it easier to hold long-lived references to an open stream.
    
1. Manual committing of messages is done by calling `mark_done()` on an open stream in read mode. To do this, you can commit the optional metadata when finished processing a message. This is similar to what's being done in adc-streaming, but since it uses confluent-kafka's `Message` type to commit messages, instead it'll hold a reference to this `Message` in an internal attributre in the `Metadata` produced that can be returned when using the Consumer..
    
2. One pattern I've seen within different projects now is to hold a long-lived reference to an open stream instance within a class, and call it occasionally. This is currently not possible for consumers since we previously just defined a function with the `@contextmanager` decorator. Instead, I reworked this a bit to not directly subclass adc-streaming's Producer and Consumer, but create new classes which hold internal references to these and define custom `__enter__` and `__exit__` methods in the internal Producer and Consumer classes. This makes it possible to hold a reference as long as you explicitly call `close()` when you're done. Another side benefit of not subclassing is that it'll make mocking easier when setting up unit tests since I can mock the internal client much easier while still testing the various methods/API calls.
    
3. Finally, the API was slightly modified. A new method in the Consumer, Consumer.read() (basically the same as adc-streaming's Consumer.stream() but renamed to not have it look like `s.stream()` which may be confusing) is used to read messages from a stream in addition to the iterator API (which is just `s.read()` without any kwargs). One can pass in the metadata kwarg here instead of stream.open() instead which is more explicit than when passing it into `stream.open()`.

Some examples:

Manual committing:
```
with stream.open("kafka://localhost:9092/my-topic", "r") as s:
    for message, meta in s.read(autocommit=False, metadata=True):
        print(message, meta)
        s.mark_done(meta)
```

No context manager:
```
s = stream.open("kafka://localhost:9092/my-topic", "r")
for message in s:
    print(s)
s.close()
```

I don't have unit tests or updated documentation yet, but wanted @spenczar's opinion on this before I take it any further. Specifically, if these changes are generally reasonable and if there are any aspects that could be improved, etc.

## Checklist

### DIFF
--- a/doc/user/stream.rst
+++ b/doc/user/stream.rst
@@ -46,7 +46,10 @@ A complete list of configurable options in :code:`Stream` are:
 * :code:`start_at`: The message offset to start at, by passing in an :code:`io.StartPosition`
 * :code:`persist`: Whether to keep a long-live connection to the client beyond EOS
 
-In addition, :code:`stream.open` provides an option to retrieve Kafka message metadata as well
+So far, all examples have shown the iterator interface for reading messages from an open
+stream. But one can instead call :code:`s.read()` directly or in the case of more specialized
+workflows, may make use of extra keyword arguments to configure an open stream. For example,
+the :code:`metadata` option allows one to retrieve Kafka message metadata as well
 as the message itself, such as the Kafka topic, key, timestamp and offset. This may
 be useful in the case of listening to multiple topics at once:
 
@@ -54,6 +57,44 @@ be useful in the case of listening to multiple topics at once:
 
     from hop import stream
 
-    with stream.open("kafka://hostname:port/topic1,topic2", "r", metadata=True) as s:
-        for message, metadata in s:
+    with stream.open("kafka://hostname:port/topic1,topic2", "r") as s:
+        for message, metadata in s.read(metadata=True):
              print(message, metadata.topic)
+
+Anatomy of a Kafka URL
+-----------------------
+
+Both the CLI and python API take a URL that describes how to connect to various
+Kafka topics, and takes the form:
+
+.. code:: bash
+
+   kafka://[groupid@]broker/topic[,topic2[,...]]
+
+The broker takes the form :code:`hostname[:port]` and gives the URL to connect to a
+Kafka broker. Optionally, a :code:`groupid` is provided which is used to keep track
+of which messages have been read from a topic with a given group ID. This allows a long-lived
+process reading messages to pick up where they left off after a restart, for example.
+Finally, one can publish to a topic or subscribe to one or more topics to consume messages
+from.
+
+Committing Messages Manually
+------------------------------
+
+By default, messages that are read in by the stream are marked as read immediately after
+returning them from an open stream instance for a given group ID. This is suitable for most cases,
+but some workflows have more strict fault tolerance requirements and don't want to lose
+messages in the case of a failure while processing the current message. We can instead commit
+messages after we are done processing them so that in the case of a failure, a process that is
+restarted can get the same message back and finish processing it before moving on to the next.
+This requires returning broker-specific metadata as well as assigning yourself to a specific group ID.
+A workflow to do this is shown below:
+
+.. code:: python
+
+    from hop import stream
+
+    with stream.open("kafka://mygroup@hostname:port/topic1", "r") as s:
+        for message, metadata in s.read(metadata=True, autocommit=False):
+             print(message, metadata.topic)
+             s.mark_done(metadata)
--- a/doc/user/stream.rst
+++ b/doc/user/stream.rst
@@ -46,6 +46,18 @@ A complete list of configurable options in :code:`Stream` are:
 * :code:`start_at`: The message offset to start at, by passing in an :code:`io.StartPosition`
 * :code:`persist`: Whether to keep a long-live connection to the client beyond EOS
 
+One doesn't have to use the context manager protocol (:code:`with` block)
+to open up a stream as long as the stream is explicitly closed afterwards:
+
+.. code:: python
+
+    from hop import stream
+
+    s = stream.open("kafka://hostname:port/topic", "r")
+    for message in s:
+         print(message)
+    s.close()
+
 So far, all examples have shown the iterator interface for reading messages from an open
 stream. But one can instead call :code:`s.read()` directly or in the case of more specialized
 workflows, may make use of extra keyword arguments to configure an open stream. For example,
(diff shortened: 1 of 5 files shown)


hop/io.py
```diff
@@ -249,39 +270,77 @@ def unpack(message, metadata=False):
...
+            metadata: A Metadata instance containing broker-specific metadata.
+
+        """
+        tp = confluent_kafka.TopicPartition(
+            topic=metadata.topic,
+            partition=metadata.partition,
+            offset=metadata.offset + 1
+        )
+        # access Kafka consumer to store offsets
+        self._consumer._consumer.store_offsets(offsets=[tp])
```
spenczar, line 286: This doesn't seem right. We shouldn't be reaching into a private variable (the inner `_consumer`) like this. I would expect you to use `self._consumer.mark_done` ([ref](https://github.com/astronomy-commons/adc-streaming/blob/master/adc/consumer.py#L82-L88)). Does that work?

myNameIsPatrick: Yeah, I'm not too happy about that either. The workaround is due to `mark_done()` only allowing `confluent_kafka.Message` as an input which won't be passed directly to the user here and can't be instantiated either so I can't create one and pass it in. An alternative could be to allow `adc.consumer.Consumer.mark_done()` to take in `TopicPartition` as well.

spenczar: I propose including the raw message in the public `Metadata` class's attributes, then. It can be a private attribute, `_raw: confluent_kafka.Message`, for example. Would that work?

myNameIsPatrick: That sounds like a good approach. I can do some finagling so that you don't have to have two copies of all the relevant properties via a `__post_init__` method and just reference the `Message` directly.

myNameIsPatrick: Alright, that should be in now. Turns out `__post_init__` and `frozen=True` don't mix well for dataclasses without an ugly workaround, and adding the frozen property has a tiny performance hit (https://docs.python.org/3/library/dataclasses.html#frozen-instances) so I'm inclined to drop it.

myNameIsPatrick:
@spenczar, thanks for your comments so far, I have implemented those changes and also added relevant unit tests and documentation now.

doc/user/stream.rst
```diff
@@ -46,14 +46,67 @@ A complete list of configurable options in :code:`Stream` are:
...
+Committing Messages Manually
+------------------------------
+
+By default, messages that are read in by the stream are marked as read immediately after
+returning them from an open stream instance for a given group ID. This is suitable for most cases,
+but some workflows have more strict fault tolerance requirements and don't want to lose
+messages in the case of a failure while processing the current message. We can instead commit
+messages after we are done processing them so that in the case of a failure, a process that is
+restarted can get the same message back and finish processing it before moving on to the next.
+This requires returning broker-specific metadata as well as assigning yourself to a specific group ID.
```
spenczar, line 102: Why does it require assigning yourself a specific group ID?

hop/io.py
```diff
@@ -214,74 +211,129 @@ def _generate_group_id(n):
...
+    timestamp: int = field(init=False)
+    key: Union[str, bytes] = field(init=False)
+    _raw: confluent_kafka.Message
+
+    def __post_init__(self):
+        self.topic = self._raw.topic()
+        self.partition = self._raw.partition()
+        self.offset = self._raw.offset()
+        self.timestamp = self._raw.timestamp()[1]
+        self.key = self._raw.key()
```
spenczar, line 231: This works, of course, but I would have expected a classmethod constructor I think:
```python
@classmethod
def from_message(cls, msg: confluent_kafka.Message) -> Metadata:
    return Metadata(
        topic=msg.topic(),
        partition=msg.partition(),
        offset=msg.offset(),
        timestamp=msg.timestamp()[1],
        key=msg.key(),
        _raw=msg,
    )
```

Obviously, just a matter of taste, but I like that you don't need the noisy `field(init=False)` annotations, and constructing things as `Metadata.from_message(msg)` seems tidy.

tests/test_io.py
```diff
@@ -73,13 +73,18 @@ def test_deserialize(message, message_parameters_dict, caplog):
...
-        broker_url1 = "kafka://hostname:port/gcn"
-        broker_url2 = "kafka://hey@hostname:port/gcn"
-        start_at = io.StartPosition.EARLIEST
+    mock_adc_consumer = mock_consumer(mock_broker, topic, group_id, start_at)
+    with patch("hop.io.consumer.Consumer", autospec=True) as mock_instance:
+        mock_instance.side_effect = mock_adc_consumer
+        mock_adc_consumer.stream.return_value = [{'hey', 'you'}]
+
+        broker_url1 = "kafka://hostname:port/{topic}"
+        broker_url2 = "kafka://{group_id}@hostname:port/{topic}"
```
spenczar, line 87: Should these be `f`-strings?

myNameIsPatrick: Definitely. This is now fixed, nice catch.

hop/io.py
myNameIsPatrick: Yup, I totally agree with that, it looks much nicer. I had to change the return annotation to the `'Metadata'` string to get this to work, otherwise get a `NameError: name 'Metadata' is not defined` error. Have you encountered that issue when adding type annotations to adc-streaming?

doc/user/stream.rst
myNameIsPatrick: If the group ID isn't specified in the URL, a random one is assigned instead. I suppose someone could reuse that one if it was directly specified upon a second invocation.

hop/io.py
spenczar: Ah yep. The reason is that there's no forward definitions in type annotations - the python interpreter walks through the file in one pass, so it can't refer to things that come later in the file, or do anything self-referential.

[PEP-563 -- Postponed Evaluation of Annotations](https://www.python.org/dev/peps/pep-0563/) proposes ways to solve this problem. In Python 3.7+, you can use `from __future__ import annotations`. In Python 3.10, it'll just work. In Python 3.6-, you're out of luck, and simply must use strings.

doc/user/stream.rst
spenczar: Right, that makes sense. You _could_ use a random group ID, but you'd probably never want to.

myNameIsPatrick:
Thanks @spenczar! I'll leave it be for now and add it to the stack of items to change once we drop python 3.6.

PR #104, opened 2020-09-16
Return message content rather than Blob when unpacking messages

myNameIsPatrick:
## Description

This PR changes the behavior of unstructured (Blob) messages when they're unpacked to return the message content rather than a Blob class to be a bit more intuitive, first noted by @shereenElSayed. To clarify, if you try to publish unstructured messages and subscribe to them, you'll get back a Blob rather than the message you originally sent which can be a bit confusing.

This also updates the tests to reflect the change and makes a few changes within the Blob class. Removed the comment about the reason for adding the Blob in the docstring, updated the type annotation for the content attribute to reflect any JSON-serializable content, and updated the `load()` method to update the content tag rather than unpacking the content into the blob (if it worked before, it is by coincidence).

## Checklist

### DIFF
--- a/hop/io.py
+++ b/hop/io.py
@@ -173,7 +173,7 @@ class _DeserializerMixin:
             raise ValueError("Message is incorrectly formatted")
         else:
             if format == Deserializer.BLOB.name:
-                return cls[format].value(content=content)
+                return content
             elif format in cls.__members__:
                 return cls[format].value(**content)
             else:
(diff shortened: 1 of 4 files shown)


bfc5288 (approved):
This looks good. I think the changes to `BLOB` are sensible -- they reflect the fact that `BLOB` is supposed to be a non-formatted message, rather than a full format of its own (which the initial implementation didn't follow, since I designed it and its tests to be symmetric with the actual message formats).

Feel free to merge when convenient.

myNameIsPatrick:
Thanks for the review, @bfc5288!

PR #109, opened 2020-09-29
Improve test reliability

### LINKED ISSUE #108: Auth tests depend on developer's home directory
### Description
As a developer, I want the unit tests to pass regardless of the state of my home directory.

`tests/test_auth.py::test_load_auth` and `tests/test_auth.py::test_setup_auth` do not set or change `XDG_CONFIG_HOME`, causing them to attempt to test against the developer's normal config file location. As a result, the tests fail if an actual, non-test config file already exists. 

### Definition of Done
Test pass when ~/.config/hop/config.toml already exists

### Tasks/Items:
- [ ] Create a temporary directory for configuration during tests and set `XDG_CONFIG_HOME` to refer to it

(The task above seems like the simplest solution, but I am unsure how this would be most idiomatically handled in the context of pytest (or whether another method would be preferred).

cnweaver:
Perform auth tests in temporary directories to avoid state leaking in
from the user's home directory. Fine-tune the version requirement to
avoid the mock_file.read() bug.

## Description

Fixes #108 

## Checklist

* [ ] ~Add/update sphinx documentation with any relevant changes.~
* [ ] ~Add/update pytest-style tests in `/tests`, ensuring sufficient code coverage.~
* [ ] ~`make doc` runs without errors and generated docs render correctly.~

### DIFF
--- a/hop/auth.py
+++ b/hop/auth.py
@@ -34,7 +34,7 @@ class Auth(auth.SASLAuth):
         super().__init__(user, password, ssl=ssl, method=method, **kwargs)
 
 
-def load_auth(config_file=configure.get_config_path()):
+def load_auth(config_file=None):
     """Configures an Auth instance given a configuration file.
 
     Args:
@@ -48,6 +48,8 @@ def load_auth(config_file=configure.get_config_path()):
         KeyError: An error occurred parsing the configuration file.
 
     """
+    if config_file is None:
+        config_file = configure.get_config_path()
     if not os.path.exists(config_file):
         raise FileNotFoundError(errno.ENOENT, os.strerror(errno.ENOENT), config_file)
 
--- a/tests/test_auth.py
+++ b/tests/test_auth.py
@@ -1,4 +1,6 @@
 from unittest.mock import patch, mock_open
+from tempfile import TemporaryDirectory
+from contextlib import contextmanager
 
 import pytest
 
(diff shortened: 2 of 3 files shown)


myNameIsPatrick (commented):
This generally looks good, always happy to see improvements in the test suite. One thing I noticed was that the lint step was ringing up: https://github.com/scimma/hop-client/pull/109/checks?check_run_id=1184006358. Out of scope for this PR, but it would be good to split off the linting job into its own check so that it's much more visible and would raise warnings or fail if there are linting issues.

cnweaver:
In fact, there is a logic bug that one of the linter wanrings has coincidentally pointed me to, so I will be fixing that, and then the rest of the linter's concerns. `make lint` was unfortunately not helpful for this, because it tries to run on the entire contents of my virtualenv and dies on something it finds in the toml package before actually checking any of the code I worked on.

myNameIsPatrick (approved):
Looks great, @cnweaver! I approve all the changes.

> make lint was unfortunately not helpful for this, because it tries to run on the entire contents of my virtualenv and dies on something it finds in the toml package before actually checking any of the code I worked on.

Ah, that's disappointing. I've never made a virtualenv in the same directory as a git repo, but I imagine that'd run into trouble as it recursively looks for any *.py files.

PR #112, opened 2020-09-30
Feature/workflow_cleanup

bfc5288:
## Description

This brief PR makes two small fixes for the CI upload workflows:

* Codecov: our workflow uses the [Codecov Action](https://github.com/codecov/codecov-action), which was recently updated to workaround Codecov's transient server timeouts (which we experienced [here](https://github.com/scimma/hop-client/runs/1184006334?check_suite_focus=true)). I specified their new version into our workflow to hopefully avoid the issue for now (but we should later revert the version line back to `uses: codecov/codecov-action@v1` so we can get later updates to their Action)
  * Another option is to not have the CI fail if there's an error, e.g. specify `fail_ci_if_error: false`
* Conda: fixed a formatting issue in the workflow that caused a build error.

* Sidenote on `conda`: currently, the `meta.yaml` conda recipe works only when invoked by our CI (which injects the version string before building), so using `make conda-build` fails locally since `version` is missing without our CI. I attempted fixing this using the [recommended jinja templating](https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#package-version) and [another workaround](https://github.com/conda/conda-build/issues/2697#issuecomment-367732506), but this still fails since the `setup.py` metadata appears to have everything except for `version`:
```
# meta.yaml
{% set setup_data = load_setup_py_data() %}
...
  keys: {{ setup_data.keys() }}
...
# referencing `keys` yields:
dict_keys(['name', 'description', 'long_description', 'long_description_content_type', 'url', 'author', 'author_email', 'license', 'packages', 'entry_points', 'python_requires', 'install_requires', 'extras_require', 'setup_requires', 'zip_safe', 'use_scm_version', 'classifiers'])
# note the lack of `version`
```
I'd like to fix the issue in this cleanup PR if possible, but don't have the solution at the moment. 

## Checklist
~~* [ ] All new functions and classes are documented and adhere to Google doc style (3.8.3-3.8.6 of [this](http://google.github.io/styleguide/pyguide.html#383-functions-and-methods) document)~~
~~* [ ] Add/update sphinx documentation with any relevant changes.~~
~~* [ ] Add/update pytest-style tests in `/tests`, ensuring sufficient code coverage.~~
~~* [ ] `make test` runs without errors.~~
~~* [ ] `make lint` doesn't give any warnings.~~
~~* [ ] `make format` doesn't give any code formatting suggestions.~~
~~* [ ] `make doc` runs without errors and generated docs render correctly.~~


myNameIsPatrick (approved):
Looks good to me. I don't have any ideas about the conda issues either, but it may be worth seriously considering moving to conda-forge rather than maintain our own conda packages.

PR #116, opened 2020-10-01
Add tests for the main functions in hop.configure

cnweaver:
## Description

The coverage checker still complains about hop/configure:94-98, although that code is really part of the CLI and is already exercised by that set of tests. However, if it would be preferable to convince the coverage tool, I can add some direct tests for `configure._main` as well. 

## Checklist

* ~[ ] All new functions and classes are documented and adhere to Google doc style (3.8.3-3.8.6 of [this](http://google.github.io/styleguide/pyguide.html#383-functions-and-methods) document)~
* ~[ ] Add/update sphinx documentation with any relevant changes.~
* ~[ ] `make doc` runs without errors and generated docs render correctly.~

### DIFF
--- a/tests/test_configure.py
+++ b/tests/test_configure.py
@@ -1,0 +1,51 @@
+import pytest
+import os
+from unittest.mock import patch, MagicMock
+
+from hop import configure
+
+
+def check_config_file(config_path, username, password):
+    assert os.path.exists(config_path)
+    cf = open(config_path, "r")
+    config_file_text = cf.read()
+    assert username in config_file_text
+    assert password in config_file_text
+
+
+def test_write_config_file(tmpdir):
+    config_file = tmpdir + "/config"
+    username = "scimma"
+    password = "scimmapass"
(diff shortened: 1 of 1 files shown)


myNameIsPatrick (approved):
Looks great! Feel free to merge this in whenever you get a chance.

> The coverage checker still complains about hop/configure:94-98, although that code is really part of the CLI and is already exercised by that set of tests. However, if it would be preferable to convince the coverage tool, I can add some direct tests for configure._main as well.

I'm guessing the blind spot is in `test_cli.py` since it's not exercising `hop configure setup` even though the underlying functions it relies on are being tested.

cnweaver:
`test_cli.py` doesn't cover that but `test_auth.py` does. However, I now notice that it looks like `test_setup_auth` may not be getting counted due to using an explicit `subprocess.Popen` instead of `script_runner.run`. I'll try changing that and put it in its own patch if it turns out to help.

PR #119, opened 2020-10-03
Rework subscribe tests, catch a couple of message-based bugs

myNameIsPatrick:
## Description

This PR reworks the print_message tests in `test_subscribe.py` to handle the `json_dump` kwarg use case as well. In doing so, two bugs were exposed:

1. `json_dump=True` was not working correctly for unstructured messages, returning an `AttributeError` since `asdict()` was being called on it.
2. `Blob` was returning an extra level of nesting when serializing messages and so that was causing some strange behavior when running `hop subscribe`.

The first bug was due to missing test coverage in our test suite. The second bug was due to our tests not really testing the `print_message` behavior properly in `subscribe.py`. If you look at https://github.com/scimma/hop-client/compare/fix/subscribe_json_opt?expand=1#diff-57367abe096617660023e36588782227L25, you'll see that the message classes are mocked, so you're not really testing how the message classes represent messages as string, but really just a test of comparing files between `tests/data/test_data` and `tests/data/expected_data` which defeats the purpose of the test. So that has been reworked to use the actual message classes instead.

I also found out why this codec business was needed in the first place: https://github.com/scimma/hop-client/compare/fix/subscribe_json_opt?expand=1#diff-57367abe096617660023e36588782227L37. Turns out that extracting the message string from stdout in this manner was mangling the output. Instead, I leverage `StringIO.getvalue()` to access the string that gets redirected from stdout, avoiding this issue.

Finally, I regenerated the expected outputs for the various message types to what's being seen from stdout and generated extra files for the json-based outputs.

## Checklist

~~* [ ] All new functions and classes are documented and adhere to Google doc style (3.8.3-3.8.6 of [this](http://google.github.io/styleguide/pyguide.html#383-functions-and-methods) document)~~
~~* [ ] Add/update sphinx documentation with any relevant changes.~~


cnweaver (approved):
These changes make sense to me; it's definitely good to eliminate places where the tests overuse mocks and succeed trivially.

myNameIsPatrick:
Thanks for the review, @cnweaver!

bfc5288:
This is after the merge, but these are some great changes @myNameIsPatrick.

That's right about the `codec` usage -- I used it for the [first subscriber tests](https://github.com/scimma/hop-client/pull/55/files#r397932554) to deal with the weird stdout redirection like you said, but it definitely wasn't a very clean solution; it's great that this PR cleans it up with only `StringIO`.

I hadn't realized that section of the `print_message` tests was trivial, so that's a good catch as well. As a sidenote -- did you have a procedure for generating the expected output files? It may be good to document that somewhere.

PR #122, opened 2020-10-13
Change the publish CLI to treat input from stdin as JSON.

cnweaver:
This allows data output from `hop subscribe` to be fed back into
`hop publish`. It is a breaking change, so users of this interface,
if any exist, should be considered.

## Description

Aside from the interface change, which consists of a single additional call to `json.loads`, most of this patch rewrites the mock_broker and surrounding test code in order to test the new feature. I was unable to get the old approach using a combination of `MagicMock` and `functools.partialmethod` to do anything sensible (all writes kept vanishing into thin air), but it's possible I just made some mistake, and if so I would be happy to refactor this patch. 

## Checklist

* ~[ ] All new functions and classes are documented and adhere to Google doc style (3.8.3-3.8.6 of [this](http://google.github.io/styleguide/pyguide.html#383-functions-and-methods) document)~
* [x] Add/update sphinx documentation with any relevant changes. (Should the expectation that unstructured data is JSON be called out more explicitly? Current wording hints at it.)

### DIFF
--- a/doc/user/commands.rst
+++ b/doc/user/commands.rst
@@ -28,9 +28,13 @@ This command allows a user to publish various structured and unstructured messag
 
 * `RFC 822 formatted GCN circular <https://gcn.gsfc.nasa.gov/gcn3_circulars.html>`_
 * An XML formatted `GCN/VOEvent notice <https://gcn.gsfc.nasa.gov/tech_describe.html>`_
-* Unstructured messages such as byte-encoded or JSON-serializable data.
+* Unstructured messages such as JSON-serializable data.
 
 Structured messages such as GCN circulars and VOEvents are published as JSON-formatted text.
+
+Unstructured messages may be piped to this command to be published. This mode of operation
+requires JSON input with individual messages separated by newlines, and the Blob format
+(`-f BLOB`) to be selected. 
 
 .. program-output:: hop publish --help
    :nostderr:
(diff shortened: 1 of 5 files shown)


myNameIsPatrick (approved):
Apologies for the delay. This looks great, @cnweaver! Had a go locally just to test out with the python API and CLI and it looks good to me.

> most of this patch rewrites the mock_broker and surrounding test code in order to test the new feature. I was unable to get the old approach using a combination of MagicMock and functools.partialmethod to do anything sensible (all writes kept vanishing into thin air), but it's possible I just made some mistake, and if so I would be happy to refactor this patch.

I'm pretty good with the rewrite. I originally wrote that old approach as an incremental improvement over the previous approach, but it was by no means great. It's fairly likely there was something wrong with that implementation with the missing writes.

> This allows data output from hop subscribe to be fed back into
hop publish. It is a breaking change, so users of this interface,
if any exist, should be considered.

I'd imagine since it's not well advertised that we don't have many, if any users of this interface. Nonetheless, our next release with this inclusion shouldn't be a point release and should be clear in the changelog.
> Should the expectation that unstructured data is JSON be called out more explicitly? Current wording hints at it.

Probably, yes. I'll leave it to you whether you want to do it as part of this PR or raise an issue about improving the documentation. I'm happy either way.

PR #123, opened 2021-02-05
Add documentation of optional parameters which are passed down to lower level Kafka libraries

cnweaver:
## Description

This only adds documentation for existing features

## Checklist

* ~[ ] Add/update pytest-style tests in `/tests`, ensuring sufficient code coverage.~

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -51,6 +51,7 @@ extensions = [
     'sphinx.ext.autosummary',
     'sphinx.ext.autosectionlabel',
     'sphinxcontrib.programoutput',
+    'sphinx.ext.napoleon',
 ]
 
 # Add any paths that contain templates here, relative to this directory.
--- a/hop/io.py
+++ b/hop/io.py
@@ -31,11 +31,12 @@ class Stream(object):
     stream connection is opened, it will use defaults specified here.
 
     Args:
-        auth: A `bool` or `Auth` instance. Defaults to loading from `auth.load_auth()`
-            if set to True. To disable authentication, set to False.
+        auth: A `bool` or :class:`Auth <hop.auth.Auth>` instance. Defaults to
+            loading from :meth:`auth.load_auth <hop.auth.load_auth>` if set to
+            True. To disable authentication, set to False.
         start_at: The message offset to start at in read mode. Defaults to LATEST.
         persist: Whether to listen to new messages forever or stop
-             when EOS is received in read mode. Defaults to False.
+            when EOS is received in read mode. Defaults to False.
 
     """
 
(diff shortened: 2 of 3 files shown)


cnweaver:
The new documentation does not show up in the generated sphinx docs, as the `hop.io._Consumer` and `hop.io._Producer` classes do not seem to appear there at all. I'm not sure what if anything to do about this.

myNameIsPatrick:
> The new documentation does not show up in the generated sphinx docs, as the `hop.io._Consumer` and `hop.io._Producer` classes do not seem to appear there at all. I'm not sure what if anything to do about this.

The main reason for that is by default, sphinx won't show any documentation for private classes or functions. Making Consumer/Producer as part of the public API is one solution, otherwise a more creative way of documenting these may be needed.

cnweaver:
My preference would be to make these classes public, since they have interfaces we want our users to use and so we want the documentation of those interfaces to be visible. It seeems like marking the __init__ methods private (maybe spinx's [`:meta private:`](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html) can do this?) would be good, though, since we don't want users to think that they should or have to instantiate these classes themselves. Unfortunately, I can't get any of this to work; if I change the class names to no longer start with underscores or even explicitly list them in

    .. automodule:: hop.io
        :members: Stream, Producer, Consumer

sphinx still ignores them and only generates documentation for the `Stream` class.

myNameIsPatrick:
Okay, so I looked into this.The reason why they're still not showing up is because the classes themselves don't have any docstrings either so sphinx is skipping them. If you move the docstrings from `__init__` to the top level class, that should work.

I don't think there's a clean way of making `__init__` private but we could discourage their use for instantiating directly by being a bit more explicit in documenting this module and labeling the Producer/Consumer lower-level classes, similar to the requests library (https://requests.readthedocs.io/en/master/api/, https://raw.githubusercontent.com/psf/requests/master/docs/api.rst).

Before going down this route, I think it'd be a good idea to make the pack/unpack methods private. Before, since the classes themselves were private, I was less concerned but now by making them private it would hide their methods in the API documentation and discourage their use.

cnweaver:
I've shuffled around what is and isn't private, hopefully covering all of the points above. 

While doing that I also realized that the documentation output looked like a mess because we were using a style which apparently isn't supported by the default sphinx parser. Adding in an appropriate preprocessor makes it work much better.

myNameIsPatrick (approved):
Looks great, @cnweaver! Just got a chance to build the docs locally and having the napoleon preprocessor in there makes the docs look much better.

PR #125, opened 2021-02-12
Require that credential files have private permissions

### LINKED ISSUE #124: User Story issue
### Description
As a hopskotch operator, I want to prompt users to store their credentials securely to reduce risk of them being compromised.

A very simple aspect of this is that users should prevent others on shared systems from reading their credential files, and while we cannot enforce this 100% of the time, we can catch a lot of mistakes if the client library refuses to load files with unsafe permissions (and tells the user how to correct them). 

### Definition of Done
The library should refuse to read credential files with permissions which are broader than `0600`, and should fail with a clear, explanatory message.

cnweaver:
Closes #124

## Description

Refuse to open config files which have over-broad permissions (and prompt the user to fix them). Also, produce config files with safe permissions by default. 

## Checklist

### DIFF
--- a/hop/configure.py
+++ b/hop/configure.py
@@ -3,6 +3,7 @@ import logging
 import os
 import csv
 import errno
+import stat
 
 import toml
 
@@ -51,6 +52,7 @@ def write_config_file(config_file, username, password):
 
     os.makedirs(os.path.dirname(config_file), exist_ok=True)
     with open(config_file, "w") as f:
+        os.chmod(config_file, stat.S_IRUSR | stat.S_IWUSR)
         toml.dump({"auth": {"username": username, "password": password}}, f)
         logger.info(f"Generated configuration at: {config_file}")
 
--- a/hop/configure.py
+++ b/hop/configure.py
@@ -51,8 +51,8 @@ def write_config_file(config_file, username, password):
     """
 
     os.makedirs(os.path.dirname(config_file), exist_ok=True)
-    with open(config_file, "w") as f:
-        os.chmod(config_file, stat.S_IRUSR | stat.S_IWUSR)
+    fd = os.open(config_file, os.O_WRONLY | os.O_CREAT, stat.S_IRUSR | stat.S_IWUSR)
+    with open(fd, "w") as f:
         toml.dump({"auth": {"username": username, "password": password}}, f)
         logger.info(f"Generated configuration at: {config_file}")
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,9 +1,13 @@
 from collections import defaultdict
+from contextlib import contextmanager
 import io
+import os
+import stat
 from unittest.mock import MagicMock
 
 import pytest
 
+from hop import configure
 from hop import models
 from hop.io import StartPosition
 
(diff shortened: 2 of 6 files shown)


cnweaver:
This ended up touching more of the tests than I had intended, but the way they were designed around assumptions of how the functions under test were implemented always gave me a bad feeling, so unless anyone has strong objections, I prefer them in the refactored form.

hop/configure.py
```diff
@@ -51,6 +52,7 @@ def write_config_file(config_file, username, password):
 
     os.makedirs(os.path.dirname(config_file), exist_ok=True)
     with open(config_file, "w") as f:
+        os.chmod(config_file, stat.S_IRUSR | stat.S_IWUSR)
```
spenczar, line 55: It sure feels like this should happen outside the `with` body. I suppose this is fine, but it raised my eyebrows at least.

tests/conftest.py
```diff
@@ -360,3 +364,51 @@ def message_parameters_dict():
...
+
+    Args:
+        data: the data to be written to the file
+        perms: the permissions which should be set on the file.
+            The default value is to use the standard, safe permissions
+
+    Returns:
+        None
+    """
+    config_path = configure.get_config_path()
```
spenczar, line 404: It feels dangerous that this function will delete my config if I invoke it wrong (that is, without overriding `XDG_CONFIG_HOME`).

Maybe this should always write into a temporary directory, and then yield that temporary directory path?

spenczar (commented):
A few comments, but this seems good.

hop/configure.py
cnweaver: I think it can't happen earlier, as the file is not gauranteed to exist until we open it. I suppose this does leave a timing window in which someone else could open the file before we read protect it and then read the data we subsequently write, but I'm not sure there's a way to entirely prevent that.

tests/conftest.py
cnweaver: This function has no latitude to choose the location, as the code under test will always just invoke `configure.get_config_path()`. However, your point about this being dangerous if someone writes a bad test is a good one. I think we can defend against that by having this function check that `XDG_CONFIG_HOME` is at least set, as any test which is not doing so is almost certainly wrong. That could still do bad things if there is both a bad test and the user keeps `XDG_CONFIG_HOME` set in their environment; I will see if we can reasonably distinguish pytest's temporary directory paths from others, so it can check that the variable is both set and points to a temporary directory.

hop/configure.py
spenczar: Oh, I meant I'd expect it to be done after the file has been closed, at the end of the with block.

It's certainly possible to do this all-in-one-go by passing a file descriptor to `open`. You can make the fd with `os.open`, which can take the permission bits. It's just a bit hairier in general but it's possible.

tests/conftest.py
spenczar: > This function has no latitude to choose the location, as the code under test will always just invoke `configure.get_config_path()`

I was picturing flipping the flow you have in the current tests. Right now, you
 - set XDG_CONFIG_HOME to a temporary directory
 - call this function to write a config file into XDG_CONFIG_HOME
 - yield execution to the actual test, which is none the wiser that it's operating against a test config in a temp directory.
 
 But you could
 - call this function to write a config file into a temporary directory, and yield it
 - set XDG_CONFIG_HOME to whatever this function yielded
 - yield execution to the actual test, which is none the wiser that it's operating against a test config in a temp directory.
 
 Would that work?

cnweaver: That should be possible; I'll try it out. I think `tmpdir` may have to be passed into `temp_config` explicitly then, unfortunately, as pytest seems to be very tight-fisted with information about where it is (I think because it's trying to make it per-test, to support concurrent tests).

hop/configure.py
cnweaver: Right, I forgot `open(2)` could do that; it's much cleaner.

myNameIsPatrick (approved):
Beyond the changes @spenczar suggested which have been address, this all looks great. Always nice to see improvements to the test suite as well.

PR #126, opened 2021-02-19
Change how consumer group names are generated

cnweaver:
When using proper authentication, prefix the group name with the credential username, instead of the local username. When a username is not available, simply use a purely random group name.

It would be cleaner to just not use a consumer group unless requested by the user, which other Kafka tools appear to support, but confluent kafka does not permit this. 

## Checklist


myNameIsPatrick (approved):
Looks good to me, @cnweaver.

> It would be cleaner to just not use a consumer group unless requested by the user, which other Kafka tools appear to support, but confluent kafka does not permit this.

Yeah, definitely.

PR #133, opened 2021-03-24
Change the internal representation to handle multiple credentials.

cnweaver:
Two features are removed, due to incompatibility with this one:
- Consumer group IDs can no longer be specified as part of a broker URL, as the userinfo is used as a credential username.
- Multiple brokers cannot be specified in a single URL, as this creates ambiguity in selecting credentials by hostname.

Writing configuration files in the new format and providing a user interface for managing them will be implemented subsequently.

## Checklist

### DIFF
--- a/hop/auth.py
+++ b/hop/auth.py
@@ -3,6 +3,7 @@ from . import configure
 import os
 import errno
 import toml
+from collections import Mapping
 
 SASLMethod = auth.SASLMethod
 
@@ -109,7 +110,7 @@ def _interpret_auth_data(auth_data):
         KeyError: A required key is missing from a credential entry.
 
     """
-    if type(auth_data) is dict:
+    if isinstance(auth_data, Mapping):
         # upgrade old-style single dict configs to new-style list-of-dicts (with one item)
         auth_data = [auth_data]
 
--- a/hop/auth.py
+++ b/hop/auth.py
@@ -64,7 +64,22 @@ def load_auth(config_file=None):
 
     """
     if config_file is None:
-        config_file = configure.get_config_path()
+        config_file = configure.get_config_path("auth")
+        # If finding data automatically, as a fallback, look to see if the auth data is in the main
+        # config in the old style.
+        if not os.path.exists(config_file):
+            main_config_file = configure.get_config_path("general")
+            if os.path.exists(main_config_file):
+                try:
+                    return load_auth(main_config_file)
+                except KeyError:
+                    # if the main config file does not contain auth data, rewrite the error into a
+                    # complaint that the auth config file does not exist, since that is what should
+                    # be fixed
+                    raise FileNotFoundError(errno.ENOENT, os.strerror(errno.ENOENT), config_file)
+            # if config.toml does not exist, continue and complain about lack of auth.toml,
+            # which is the real issue
+
     if not os.path.exists(config_file):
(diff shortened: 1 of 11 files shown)


hop/auth.py
```diff
@@ -66,28 +91,124 @@ def load_auth(config_file=None):
...
+        auth_data: The data obtained from the top-level 'auth' TOML key
+
+    Returns:
+        A list of configured Auth instances.
+
+    Raises:
+        KeyError: A required key is missing from a credential entry.
+
+    """
+    if type(auth_data) is dict:
```
myNameIsPatrick, line 112: I wonder if this may be better if you were to check if `auth_data` it was dict-like versus checking if the type is explicitly a dict, e.g.

```
from collections import Mapping

...

if isinstance(auth_data, Mapping):
```

hop/auth.py
```diff
@@ -30,14 +32,20 @@ class Auth(auth.SASLAuth):
         to the certificate.
     """
 
-    def __init__(self, user, password, ssl=True, method=SASLMethod.SCRAM_SHA_512, **kwargs):
+    def __init__(self, user, password, host="", ssl=True, method=SASLMethod.SCRAM_SHA_512,
```
myNameIsPatrick, line 35: Does a host matching the empty string do anything special? I've seen that used in a few places.

hop/list_topics.py
```diff
@@ -45,3 +49,9 @@ def _main(args):
 
 def _add_parser_args(parser):
     cli.add_client_opts(parser)
+
```
myNameIsPatrick, line 52: Since this is just listing the possible topics the user has access to, is passing in the group ID ever needed?

myNameIsPatrick (changes requested):

hop/auth.py
cnweaver: That is what I originally wanted, but I only knew how to do it through the C interface, and gave up after a while of unsuccessful searches through the docs. In retrospect I shoud have taken a hint from the C functions all having 'Mapping' in their names. Thanks!

hop/auth.py
cnweaver: It does. `select_matching_auth` will consider a credential with an empty hostname a possible, but inexact match for any hostname. Such a credential will be selected if there is no truly exact match (and there is only one such inexact match).

hop/list_topics.py
cnweaver: It is not required by Kafka, and in fact I suspect doesn't even do anything. Unfortunately, Confluent Kafka refuses to create a consumer object without a group: https://github.com/confluentinc/confluent-kafka-python/blob/b7f8dce998ec254f54c15e514850d7404c6a71a3/src/confluent_kafka/src/confluent_kafka.c#L1851

hop/auth.py
myNameIsPatrick: Thanks for the explanation, that makes sense.

hop/list_topics.py
myNameIsPatrick: Just realized I didn't highlight the correct lines. Oh yeah, I understand that about the group ID in terms of the confluent client requiring this. I wonder if we need to expose this to the user though, since any group ID should supposedly return the same results and we could just generate a random group ID to satisfy that.

cnweaver: Good point, I don't think there's a need to expose this, so I'll remove the user-facing option.

PR #138, opened 2021-04-27
Bug/auth data fidelity

### LINKED ISSUE #137: write_auth_data loses secondary auth data
### Description

While `auth._interpret_auth_data` can read several optional per-credential fields (protocol, ssl_ca_location, mechanism) `auth.write_auth_data` does not write any of them, which causes data loss when one or more of these are set to non-default values and a credential is written to disk. 

### Definition of Done
All credential properties should survive a serialization/deserialization roundtrip. 

### Tasks/Items:
- [ ] Improve `write_auth_data` to store secondary properties, at least if they have non-default values. 
- [ ] Add round-trip tests for all secondary credential properties

cnweaver:
## Description

Write optional credential fields as appropriate.
Support reading optional fields from CSV.
Fixes #137.

## Checklist

~* [ ] Add/update sphinx documentation with any relevant changes.~

### DIFF
--- a/hop/auth.py
+++ b/hop/auth.py
@@ -48,18 +48,52 @@ class Auth(auth.SASLAuth):
 
     @property
     def username(self):
+        """The username for this credential"""
         return self._username
 
     @property
+    def password(self):
+        """The password for this credential"""
+        return self._config["sasl.password"]
+
+    @property
     def hostname(self):
+        """The hostname with which this creential is associated,
+            or the empty string if the credential did not contain this information
+        """
         return self._hostname
+
+    @property
+    def mechanism(self):
+        """The authentication mechanism to use"""
+        return self._config["sasl.mechanism"]
+
+    @property
+    def protocol(self):
+        """The communication protocol to use"""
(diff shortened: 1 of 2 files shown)


cnweaver:
One possibly non-ideal aspect of this patch is that it doesn't really distinguish whether the CA cert data location being used for a given credential is the default one or was previously explicitly specified, so I find that it is baking paths specific to a particular virtual environment into my auth config. I'm not sure how much we care, though, or what remedy we might prefer for this.

myNameIsPatrick:
Looks good to me, @cnweaver!

> One possibly non-ideal aspect of this patch is that it doesn't really distinguish whether the CA cert data location being used for a given credential is the default one or was previously explicitly specified, so I find that it is baking paths specific to a particular virtual environment into my auth config. I'm not sure how much we care, though, or what remedy we might prefer for this.

I'm not sure either, but this PR clearly is a step in the right direction so I guess we can cross that bridge when we run into problems I suppose.

myNameIsPatrick (approved):

PR #155, opened 2021-09-22
Support new publication features offered by adc.

cnweaver:
## Description

This includes message headers, and changes the delivery callback to be on a per-message basis, rather than per stream.

This depends on https://github.com/astronomy-commons/adc-streaming/commit/a3daccaa86ff072aad75402c39e094b24be690bf which is not yet in a release. 

## Checklist

* [ ] All new functions and classes are documented and adhere to Google doc style (3.8.3-3.8.6 of [this](http://google.github.io/styleguide/pyguide.html#383-functions-and-methods) document)
* [ ] Add/update sphinx documentation with any relevant changes.
* [ ] Add/update pytest-style tests in `/tests`, ensuring sufficient code coverage.
* [ ] `make test` runs without errors.
* [ ] `make lint` doesn't give any warnings.
* [ ] `make format` doesn't give any code formatting suggestions.

### DIFF
--- a/hop/io.py
+++ b/hop/io.py
@@ -70,7 +70,7 @@ class Stream(object):
         else:
             return self._auth
 
-    def open(self, url, mode="r", group_id=None):
+    def open(self, url, mode="r", group_id=None, **kwargs):
         """Opens a connection to an event stream.
 
         Args:
@@ -107,7 +107,7 @@ class Stream(object):
                 raise ValueError("must specify exactly one topic in write mode")
             if group_id is not None:
                 warnings.warn("group ID has no effect when opening a stream in write mode")
-            return Producer(broker_addresses, topics[0], auth=credential)
+            return Producer(broker_addresses, topics[0], auth=credential, **kwargs)
         elif mode == "r":
             if group_id is None:
                 username = credential.username if credential is not None else None
@@ -120,6 +120,7 @@ class Stream(object):
                 start_at=self.start_at,
                 auth=credential,
                 read_forever=not self.until_eos,
+                **kwargs,
             )
         else:
             raise ValueError("mode must be either 'w' or 'r'")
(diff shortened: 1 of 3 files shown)


myNameIsPatrick (commented):
The implementation looks fine to me. We should get a new adc-streaming release out so we can have a working set of tests to look at.

Since we are now supporting headers, it would also be good to update the message metadata (https://github.com/scimma/hop-client/blob/master/hop/io.py#L232-L254) to also include headers as well via https://docs.confluent.io/platform/current/clients/confluent-kafka-python/html/index.html#message.

It would also be good to add supporting documentation on how this would be used within the client.

doc/user/stream.rst
```diff
@@ -110,3 +110,36 @@ A workflow to do this is shown below:
         for message, metadata in s.read(metadata=True, autocommit=False):
              print(message, metadata.topic)
              s.mark_done(metadata)
+
+Attatching Metadata to Messages
```
myNameIsPatrick, line 114: Minor typo here.

doc/user/stream.rst
```diff
@@ -110,3 +110,36 @@ A workflow to do this is shown below:
...
+        s.write({"my": "message"}, headers={"priority": "1", "sender": "test"})
+        s.write({"my": "other message"}, headers=[("priority", "2"), ("sender", "test")])
+
+.. code:: python
+
+    from hop import stream
+
+    with stream.open("kafka://hostname:port/topic1", "r") as s:
+      for message, metadata in s.read(metadata=True):
+        print(message, metadata.headers)
```
myNameIsPatrick, line 145: Indentation looks a little funny here.

myNameIsPatrick (changes requested):
Looks generally good to me @cnweaver, just a few minor comments on the documentation. After that's fixed, happy to approve.

PR #159, opened 2021-09-30
UX: Improve logging in CLI when publishing/subscribing

### LINKED ISSUE #146: UX: Improve logging in CLI when publishing/subscribing
### Description

As a user, I'd like better messages being displayed so that it's clear what the expected behavior is. Some possible downfalls with the current setup is (with minimal logs):

* unclear if anything is happening when subscribing to a topic unless a message arrived.
* no feedback to user in general (e.g. ctrl+c will stop listening when `--persist` is set on subscribe)
* doesn't display what topics are being listened to

It would be nice to give some extra logging (enabled by default) that can be turned off (`-q`, `--quiet` or similar). This should work seamlessly with file redirects (stderr/stdout).

myNameIsPatrick:
## Description

Implements #146. Added some logging functionality in the cli module and shared throughout our CLI commands. By default, INFO-level logs are displayed and routed to stderr to allow piping/redirects to continue working. This can be changed with `--quiet` which only shows warnings/errors or `--verbose` which also shows DEBUG-level logs as well.

A couple of other changes unrelated to this but was caught when adding in logging include:
* Organize imports in `auth` module. I had to add an import and saw it wasn't organized like some of the other modules.
* Remove default signal handler in `__main__.py`. This was causing issues with KeyboardInterrupts and prematurely exiting the program without closing the producer/consumer instances. Instead, this case is handled explicitly now.
* In a similar spirit to the previous point, I noticed that the producer didn't call `close()` when using the context manager, so that's now being done as well.
* Some tests were modified to run with `--quiet` to test stderr/stdout in a similar manner as before. Also the test_configure test was modified to not bake in the 'usage' so changes to the command line arguments don't need a change in the test.

Example logs:

```
$ hop subscribe --no-auth kafka://localhost:9092/test -s EARLIEST -e
2021-09-29 15:59:52,785 | hop : INFO : group ID not specified, generating a random group ID: 5VSQC2J1KR
2021-09-29 15:59:52,785 | hop : INFO : connecting to kafka://localhost:9092
2021-09-29 15:59:52,785 | hop : INFO : subscribing to topics: ['test']
2021-09-29 15:59:52,788 | hop : INFO : processing messages from stream
one
two
three
four
2021-09-29 15:59:53,789 | hop : INFO : finished processing messages
2021-09-29 15:59:53,789 | hop : INFO : closing connection
```

```
$ cat test.txt | hop publish --no-auth kafka://localhost:9092/test
2021-09-29 16:00:15,652 | hop : INFO : connecting to kafka://localhost:9092
2021-09-29 16:00:15,653 | hop : INFO : publishing to topic: test
2021-09-29 16:00:15,653 | hop : INFO : publishing messages to stream
```

Closes #146.

## Checklist

### DIFF
--- a/hop/__main__.py
+++ b/hop/__main__.py
@@ -1,5 +1,5 @@
 import argparse
-import signal
+import logging
 import sys
 import traceback
 
@@ -11,6 +11,9 @@ from . import subscribe
 from . import version
 from . import list_topics
 from .utils import cli as cli_utils
+
+
+logger = logging.getLogger("hop")
 
 
 def set_up_cli():
@@ -81,9 +84,8 @@ To load your credential, run `hop auth add`
 
 
 def main():
-    signal.signal(signal.SIGINT, signal.SIG_DFL)
-
     parser = set_up_cli()
+
     try:
         args = parser.parse_args()
     except SystemExit:
@@ -91,6 +93,8 @@ def main():
         raise
     try:
         args.func(args)
+    except KeyboardInterrupt:
+        logger.info("received keyboard interrupt, closing")
     except Exception as ex:
         if args.debug:
             traceback.print_exc(file=sys.stderr)
--- a/hop/auth.py
+++ b/hop/auth.py
@@ -1,13 +1,17 @@
-from adc import auth
-from . import configure
-import os
-import getpass
-import logging
+from collections.abc import Mapping
 import csv
 import errno
+import getpass
+import logging
+import os
 import stat
 import toml
-from collections.abc import Mapping
+
+from adc import auth
+
+from . import configure
+from . import cli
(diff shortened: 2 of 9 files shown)


cnweaver (approved):
The code looks good. I think in principle we should add a CLI test that ensures that the right logging levels are used/messages are emitted at the right levels, but it seems like a low priority.

myNameIsPatrick:
> The code looks good. I think in principle we should add a CLI test that ensures that the right logging levels are used/messages are emitted at the right levels, but it seems like a low priority.

Thanks @cnweaver, that's a good idea. I'll add the tests now since I think it'll be pretty straightforward.

PR #169, opened 2021-10-18
Further adaptations of the hop publication interface for the robust publisher

cnweaver:
## Description

The main changes are the addition of `Producer.write_raw`, and promoting `Producer.pack` to be part of the public interface. In addition, `pack` now processes the headers, with an eye toward it being able to attach new metadata there in the future. A `flush` method is also added to `Producer`.

## Checklist

### DIFF
--- a/hop/io.py
+++ b/hop/io.py
@@ -407,8 +407,9 @@ class Producer:
 
         Args:
             message: The message to write.
-            headers: Any headers to attach to the message, either as a dictionary
-                mapping strings to strings, or as a list of 2-tuples of strings.
+            headers: The set of headers requested to be sent with the message, either as a
+                     mapping, or as a list of 2-tuples. In either the mapping or the list case,
+                     all header keys and values should be either string-like or bytes-like objects.
             delivery_callback: A callback which will be called when each message
                 is either delivered or permenantly fails to be delivered.
 
@@ -444,7 +445,8 @@ class Producer:
         Args:
             message: The message to pack and serialize.
             headers: The set of headers requested to be sent with the message, either as a
-                     mapping of strings to strings, or as a list of 2-tuples of strings.
+                     mapping, or as a list of 2-tuples. In either the mapping or the list case,
+                     all header keys and values should be either string-like or bytes-like objects.
 
         Returns: A tuple containing the serialized message and the collection of headers which
                  should be sent with it.


hop/io.py
```diff
@@ -396,29 +396,75 @@ def write(self, message, headers=None, delivery_callback=errors.raise_delivery_e
...
+
+        Returns: A tuple containing the serialized message and the collection of headers which
+                 should be sent with it.
 
-        :meta private:
         """
+        # canonicalize headers to list form
+        if headers is None:
+            headers = []
+        if type(headers) == dict:
```
myNameIsPatrick, line 439: The check for a dict would be better served by using `isinstance`. Also, since you don't really care if it's actually a `dict` but instead just is dict-like and exposes the `items()` method, you can do this:

```
from collections.abc import Mapping
...
if isinstance(headers, Mapping):
```

hop/io.py
```diff
@@ -396,29 +396,75 @@ def write(self, message, headers=None, delivery_callback=errors.raise_delivery_e
...
+            headers = list(headers.items())
+
+        # ensure all headers are encoded
+        def ensure_bytes_like(thing):
+            try:  # check whether thing is bytes-like
+                memoryview(thing)
+                return thing  # keep as-is
+            except TypeError:
+                return thing.encode("utf-8")
```
myNameIsPatrick, line 448: Personally, I think adding the function inline is less readable than if we moved this to a private function in the same module.

Regarding the content of the function, how about adding an `else` to the try/except?
```
            try:  # check whether thing is bytes-like
                memoryview(thing)
            except TypeError:
                return thing.encode("utf-8")
            else:
                return thing 
```

hop/io.py
```diff
@@ -396,29 +396,75 @@ def write(self, message, headers=None, delivery_callback=errors.raise_delivery_e
...
 
-        :meta private:
         """
+        # canonicalize headers to list form
+        if headers is None:
+            headers = []
+        if type(headers) == dict:
+            headers = list(headers.items())
+
+        # ensure all headers are encoded
```
myNameIsPatrick, line 442: I'm a bit confused why the method signature asks for strings if we're trying to ensure this is bytes-like. Is this to make the interface friendlier to users or is there a different reason I'm missing?

hop/io.py
```diff
@@ -396,29 +396,75 @@ def write(self, message, headers=None, delivery_callback=errors.raise_delivery_e
...
+        def ensure_bytes_like(thing):
+            try:  # check whether thing is bytes-like
+                memoryview(thing)
+                return thing  # keep as-is
+            except TypeError:
+                return thing.encode("utf-8")
+        for i, header in enumerate(headers):
+            headers[i] = (ensure_bytes_like(header[0]),
+                          ensure_bytes_like(header[1]))
```
myNameIsPatrick, line 451: What about a list comprehension instead?
```
headers = [(ensure_bytes_like(k), ensure_bytes_like(v)) for k, v in headers]
```

tests/conftest.py
```diff
@@ -199,10 +199,10 @@ def reset(self):
         self._messages = defaultdict(list)
         self._offsets = defaultdict(dict)
 
-    def write(self, topic, msg, headers=None):
+    def write(self, topic, msg, headers=[]):
```
myNameIsPatrick, line 202: For this and the subsequent lines, I'd avoid sticking a mutable item like a list in function/method signatures as it could lead to hard-to-debug issues.

hop/io.py
cnweaver: Yes, accepting strings is friendlier to users, although perhaps the documentation needs to note that bytes-like data is also acceptable.

hop/io.py
cnweaver: I personally find it more readable not to have things like this floating around at the module scope when they're only intended for use in one place, but I also don't care strongly.

tests/conftest.py
cnweaver: If you feel strongly I can change it back; doing so requires making the method body more complex to replace the `None` with an empty list to avoid violating the expected type invariant. I was viewing this as a 'if it hurts when you do that, don't do that' kind of thing, given the single line method body. Since this was test code it's also less careful/tolerant of bad input than the actual implementation, but that could be changed.

Issue #157, opened 2021-09-23
Consider making the credential domain specification more robust


aepace:
When running `hop auth add`, the user can optionally enter the hostname, which is useful for juggling multiple credentials across multiple hosts. However, the hostname matching in the client is not robust against host formatting specifications. 

### Scenario 1:

User gets prompted to optionally enter the hostname when adding a new credential. User enters `kafka://kafka.scimma.org`, since that's the URL that gets entered when interacting with the `hop-client` API. The resultant entry in `auth.toml` looks like:

```
[[auth]]
username = "alexander.pace-ecbf9436"
password = "-top-secret-"
protocol = "SASL_SSL"
mechanism = "SCRAM-SHA-512"
hostname = "kafka://kafka.scimma.org"
ssl_ca_location = "/path/to/lib/python3.8/site-packages/certifi/cacert.pem"
```

Then interacting with the client code results in:

```
In [1]: from hop.list_topics import list_topics

In [2]: list_topics('kafka://kafka.scimma.org')
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
<ipython-input-2-6aac4c97738a> in <module>
----> 1 list_topics('kafka://kafka.scimma.org')

~/LIGO/venvs/py38-dev-env/lib/python3.8/site-packages/hop/io.py in list_topics(url, auth)
    450     if auth is True:
    451         credentials = load_auth()
--> 452         user_auth = select_matching_auth(credentials, broker_addresses[0], username)
    453     elif auth is not False:
    454         user_auth = auth

~/LIGO/venvs/py38-dev-env/lib/python3.8/site-packages/hop/auth.py in select_matching_auth(creds, hostname, username)
    288         if username is not None:
    289             err += f" with username '{username}'"
--> 290         raise RuntimeError(err)
    291
    292     if len(matches) > 1:

RuntimeError: No matching credential found for hostname 'kafka.scimma.org'
```

### Scenario 2:

User learned their lesson after the first attempt and omitted the `kafka://` scheme from the hostname entry. The resultant entry in `auth.toml` looks like:

```
[[auth]]
username = "alexander.pace-ecbf9436"
password = "-top-secret-"
protocol = "SASL_SSL"
mechanism = "SCRAM-SHA-512"
hostname = "kafka.scimma.org"
ssl_ca_location = "/path/to/lib/python3.8/site-packages/certifi/cacert.pem"
```
This time, the hostname look up works:

```
[...]

PR #170, opened 2021-10-19
Accept but ignore `kafka://` prefixes on manually entered hostnames

cnweaver:
## Description

Addresses https://github.com/scimma/hop-client/issues/157 .
Since we are not actually asking for a URL, and URL parsing does surprising things when no scheme is present, this change instead tries to be minimally clever and just uses a regular expression which accepts a narrow set of (hopefully) definitely good inputs, and just complains to the user about anything it can't understand. 
It would be easy enough to accept (and discard) arbitrary URL schemes if we wanted to; for now this has not been done on the groups that anyone entering a non-`kafka:` scheme is likely to be on the wrong track entirely. 
There may also be room to improve the error message/help text. 

## Checklist

### DIFF
--- a/hop/auth.py
+++ b/hop/auth.py
@@ -309,6 +309,34 @@ def select_matching_auth(creds, hostname, username=None):
     return matches[0]
 
 
+def _validate_hostname(input: str):
+    """Ensure that a value entered for a hostname looks plausible.
+
+    The user is supposed to enter a hostname, possibly with a port, not a URL.
+    However, since users are likely to enter something valid prefixed with a `kafka://` scheme,
+    accept that with a warning.
+    We do not use standard URL parsing since the input is not supposed to be a URL, and attempting
+    to parse it as such when no scheme is present mostly just leads to ambiguous results.
+
+    Args:
+        input: A string entered by the user when prompted for a hostname.
+
+    Returns:
+        The sanitized hostname.
+
+    Raises:
+        RuntimeError: If the user input is not acceptable.
+    """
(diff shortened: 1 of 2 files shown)


hop/auth.py
```diff
@@ -336,6 +337,15 @@ def read_new_credential(csv_file=None):
...
         hostname = input("Hostname (may be empty): ")
+        # check that the user entered a sensible hostname, possibly with a port
+        name_re = re.compile("^(kafka://)?([^:/]*(:[0-9]*)?)$")
+        match = name_re.match(hostname)
+        if match is None:
+            raise RuntimeError("Unable to parse hostname. "
+                               "Please enter either `hostname` or `hostname:port`.")
+        if match.group(1) is not None:
+            logger.info(f"Ignoring '{match.group(1)}' prefix on hostname")
+        hostname = match.group(2)
```
myNameIsPatrick, line 348: It may be better for this to be extracted into a separate function where the validation can be tested in isolation. It seems like `read_new_credential` is doing a bit too much now. The input sanitation may also be of use in other places as well, for example if a credential is not added via `hop auth add` but instead one of the existing credentials are modified.

myNameIsPatrick (changes requested):

hop/auth.py
cnweaver: Done. This isn't currently applied to data read from a CSV file, on the grounds that we're supposed to generate those with valid data, aand there's not any way to modify a credential, but if we add one this can be used for it.

myNameIsPatrick: That sounds reasonable to me.

myNameIsPatrick (approved):

PR #177, opened 2022-02-08
Apache Avro support

cnweaver:
## Description

This patch is intended to refactor the message modelss, their interface, the storage format in Kafka, add support for Avro messages, and support opaque binary messages (including messages produced by other clients). 

## Checklist


cnweaver:
The message model interface is not yet finalized and tests are not yet complete

lpsinger:
I am concerned that this Avro implementation is not protocol-compatible with the Confluent Wire Format adopted by LSST.

cnweaver:
We are actively investigating this same concern. It may be technically infeasible, however, to implement/enforce use of the Confluent Wire Format, which requires both a schema registry and that users use only registered schemas. We're following up with LSST on how they have implemented it and plan to use it.

lpsinger:
I do like the new way of identifying the format in the message header rather than in a JSON envelope. Would it be possible to split that out as a separate PR and get that merged?

Note, I would prefer that the header be given a namespace prefix so that it is called `x-hop-format` or something similar.

manning-ncsa:
Although my understanding of the code is insufficient to render conclusive judgement on this merge, my review if the code and the conversation with @cnweaver  lead me to believe there is nothing obviously wrong with this update. I like the support for arbitrary formats that can be utilized in the future, and the fact that we are no longer assuming JSON for all messages.

lpsinger:
I stand by my concern that this Avro implementation will create confusion because it is not protocol-compatible with the Confluent Wire Format adopted by LSST. I would prefer to see the format metadata change split out into a separate PR.

cnweaver:
@lpsinger Do you think it would be useful for controlling confusion for the documentation to spell out that the [official Avro container format](https://avro.apache.org/docs/current/spec.html#Object+Container+Files) is used in this specific message model, not the Avro variant of the [Confluent wire format](https://docs.confluent.io/platform/current/schema-registry/serdes-develop/index.html#wire-format)? As long as we support multiple formats, I don't think that the fact that VRO is producing data in one way means that no one else should be allowed to use some of the same technologies for building blocks of other things, instead we should help clarify what is used when/where. Otherwise, it would feel a bit like forbidding people from using Zlib to compress other things because it happens to be used as an implementation detail in the PNG format.

cmessick:
O

cmessick:
I just want to voice my support for this in some official capacity. I have tested this with the LVK low latency alert pipeline and have successfully sent and consumed avro messages.

cnweaver:
I think it's fair to say that there are some use cases, such as for high rates of small messages, in which embedded schemas as unsuitable. However, I don't agree that this is never appropriate, just because the data is going to pass through Kafka, and I think there's value in letting users choose to try out different strategies as they see fit. In the interest of not letting a perfect solution that we haven't fully specified further delay a solution which is at least no worse than what we had before, and I think in some ways usefully better, I am going to proceed with merging this now. There's still room to build more things that will support additional uses as we go along.

lpsinger:
> I just want to voice my support for this in some official capacity.

I think we should be very clear here that neither @cmessick nor I are speaking in any "official" capacity at all.

cmessick:
> > I just want to voice my support for this in some official capacity.
> 
> I think we should be very clear here that neither @cmessick nor I are speaking in any "official" capacity at all.

Just confirming this, we're both here as individuals, and don't represent anyone other than ourselves.

PR #190, opened 2022-08-31
Allow AvroBlob to take either a single record or a list of records

cmessick:
## Description

I've had requests from consumers of streams that basically boil down to wanting the AvroBlob and JSONBlob objects to be capable of taking the same form of input, i.e. changing AvroBlob so that users can pass the same record to either message model and  can access the content from the deserialized message in the exact same way.

This PR accomplishes this by adding a new kwarg to several of the AvroBlob methods, `single_record`, which is True by default. If it is left as True, AvroBlob will expect a record not packed in a list everywhere, just like JSONBlob. If set to false, then AvroBlob can handle a list of an arbitrary number of records, like it does currently. I haven't updated the documentation yet because I wanted to make sure that this isn't a crazy idea.

## Checklist

* [ ] Add/update sphinx documentation with any relevant changes.
* [x] `make lint` doesn't give any warnings. (this does produce warnings but they're not related to my changes as far as I can tell)
                                                                                                                                                                                                                   
### DIFF
--- a/hop/models.py
+++ b/hop/models.py
@@ -263,15 +263,20 @@ class AvroBlob(MessageModel):
 
     """
 
-    content: List[JSONType]  # serializing as Avro supports essentially the same types as JSON
+    # serializing as Avro supports essentially the same types as JSON
+    content: Union[JSONType, List[JSONType]]
     schema: dict = None
     format_name = "avro"
+    single_record: bool = True
 
-    def __init__(self, content: List[JSONType], schema: dict = None):
-        if not isinstance(content, collections.abc.Sequence):
-            raise TypeError("AvroBlob requires content to be a sequence of records")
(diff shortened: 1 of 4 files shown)


lpsinger:
We have a FIXME comment at https://git.ligo.org/emfollow/gwcelery/-/blob/v2.5.1/gwcelery/tasks/alerts.py?ref_type=tags#L217-218:

```python
    # FIXME Drop get_payload_content method once
    # https://github.com/scimma/hop-client/pull/190 is merged
```

Would you please review this PR?

cnweaver:
Yes, sorry this got neglected for so long. The basic idea looks good, and I will try to get it taken care of in the next couple of days.

cnweaver:
Since this was delayed by my not noticing it, I just added the documentation (and a little test to cover the one code path that didn't have coverage). Otherwise, all changes proposed in this PR went into #220 , so I'll close this one.

PR #192, opened 2022-10-25
Implement client-assigned message identifiers

cnweaver:
## Description

This patch implements random UUID message identifiers. The producer attaches these to messages, but the consumer does not (and is not expected to) do anything with them, besides including them in the full set of message headers. 

## Checklist

~* [ ] All new functions and classes are documented and adhere to Google doc style (3.8.3-3.8.6 of [this](http://google.github.io/styleguide/pyguide.html#383-functions-and-methods) document)~

### DIFF
--- a/doc/user/stream.rst
+++ b/doc/user/stream.rst
@@ -143,3 +143,26 @@ shown below:
     with stream.open("kafka://hostname:port/topic1", "r") as s:
         for message, metadata in s.read(metadata=True):
             print(message, metadata.headers)
+
+Standard Headers
+----------------
+
+The Hop client produces and uses certain message headers automatically. It is designed so that each
+header is intended to be optional, in the sense that messages lacking these headers can still be
+processed, but if a header is missing, functionality based on it may not be available. Headers
+currently automatically produced and used are:
+
+* :code:`_id`: The value of this header is a unique ID intended to allow referring to the specific
+  message without requiring context like its position within a Kafka topic. Message IDs are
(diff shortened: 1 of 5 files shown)


rdt12 (approved):
Hi Chris, one thing that would be nice is some sort of requirements document or issue number that the MR could refer to document why the headers were added.

There's nothing technically wrong with the changes in the MR, but it's not clear what problem they are trying to solve and so it's not possible to say whether they correctly solve a given problem or not.

For example, the `_sender` header is definitely added. However, it's added client side so it wouldn't meet the needs of anyone that wanted to be absolutely sure that `_sender=foo` absolutely implies that foo authenticated with the server and sent the message. It does meet the needs of someone that is ok with the data being "mostly reliable" (think SMTP `From` and `Received` headers. 

For the `_id` header, it's not clear what problem a random UUID solves. I know that this was discussed on zoom in one of the sprint meetings, but it would be nice to have the results of the discussion written somewhere and referenced in the MR.

The above being said, I didn't see any technical problems and the changes do implement what is described in the MR description so I am approving.

cnweaver:
Fair point about the reasoning behind these changes. I agree that this is the sort of thing that I think would ideally be in a design/requirements document, but we don't really write or maintain any of those. It does seem like I should update the docs to indicate the level of trust associated with this data, since even end users might need to be aware of that. It feels a little odd to retro-actively make an issue, so I'll add some description of our reasoning to the description here.

PR #198, opened 2023-01-07
Create Blob instead of crashing on binary data on deserialize

alchzh:
## Description

Currently, reading a message that has binary data without the proper deserialize plugin throws a UnicodeDecodeError exception.

I think the expected behavior is to create a Blob with the bytes.

This PR ignores any UnicodeDecodeErrors alongside JSONDecodeErrors

## Checklist

* [ ] `make lint` doesn't give any warnings.
* [ ] `make format` doesn't give any code formatting suggestions.

### DIFF
--- a/hop/io.py
+++ b/hop/io.py
@@ -218,7 +218,7 @@ class _DeserializerMixin:
                 return from_format(old["content"], old["format"], deserialize=False)
             # not labeled according to our scheme, but it is valid JSON
             return models.JSONBlob(content=old)
-        except json.JSONDecodeError:  # if we can't tell what the data is, pass it on unchanged
+        except (UnicodeDecodeError, json.JSONDecodeError):  # if we can't tell what the data is, pass it on unchanged
             logger.warning("Unknown message format; returning a Blob")
             return models.Blob(content=message.value())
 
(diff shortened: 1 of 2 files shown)


alchzh:
I've corrected the test case that let this slip through (\x00 is valid unicode decodable)

It seems like a lint and format is required, but not because of this change.

hop/io.py
```diff
@@ -218,7 +218,7 @@ def from_format(data, format, deserialize=True):
                 return from_format(old["content"], old["format"], deserialize=False)
             # not labeled according to our scheme, but it is valid JSON
             return models.JSONBlob(content=old)
-        except json.JSONDecodeError:  # if we can't tell what the data is, pass it on unchanged
+        except (UnicodeDecodeError, json.JSONDecodeError):  # if we can't tell what the data is, pass it on unchanged
```
cnweaver, line 221: The linter becomes unhappy about the length of this line; the comment can just be moved down onto a line of its own to fix that.

cnweaver (changes requested):
Thanks for catching this bug, and providing a nice fix and updated test. 

Sorry about the mess with the linter; I keep forgetting to fix that because I don't want to mix it with other changes. I've been kicking the can down the road by checking that the linter doesn't produce _new_ complaints, however, I will get this cleaned up ASAP.

cnweaver (approved):

alchzh:
Found out that on Python 3.6+ `json.loads` accepts `bytes` as an argument anyway, so we can just skip the `.decode()` and second exception type entirely.

It only makes a difference if the json isn't encoded as `utf-8` (see the encoding detector here https://github.com/python/cpython/blob/3.11/Lib/json/__init__.py#L244)

PR #206, opened 2023-10-10
Fix authentication procedure in RobustPublisher

alchzh:
## Description

RobustPublisher currently does not work at all because it's using an older auth API. It isn't caught in the tests because `Stream` gets mocked like the old format. This PR fixes it using the credential selection code from `Stream.open`.

## Checklist


cnweaver (approved):
Thanks for both catching this and providing a fix!

PR #209, opened 2024-03-14
Expose stop method for Consumer

myNameIsPatrick:
## Description

Exposes the stop method for the underlying adc-streaming Consumer class.

## Checklist

### DIFF
--- a/hop/io.py
+++ b/hop/io.py
@@ -401,6 +401,12 @@ class Consumer:
         """
         self._consumer.mark_done(metadata._raw, asynchronous)
 
+    def stop(self):
+        """Stops the runloop of the consumer. Useful when running the
+        consumer in a different thread.
+        """
+        self._consumer.stop()
+
     def close(self):
         """End all subscriptions and shut down.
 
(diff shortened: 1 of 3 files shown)


myNameIsPatrick:
Note: I could add test coverage using a mock, but this won't sufficiently test the method in question as the underlying adc-streaming Consumer class is being mocked in all tests. Any suggestions?

cnweaver:
I think that at the level of a hop-client-specific unit test it can be considered sufficient to test that the method does in fact pass through to the lower-level (adc) consumer; after all what we're really expecting is that adc-streaming will also just pass through to confluent-kafka-python, and that that will pass the call through to librdkafka, which will actually do the right thing. It doesn't seem crazy to verify that the new code correctly forms the newest link in that chain, rather than worrying about trying to test whether the implementation buried in librdkafka really does stop its internal loop.

myNameIsPatrick:
That's good enough for me. I've added the relevant test.

cnweaver (approved):
LGTM!

PR #211, opened 2024-06-12
Expose the capability to multiple topics

cnweaver:
## Description

Based on https://github.com/astronomy-commons/adc-streaming/pull/69 , this exposes the same functionality through the hop API. 

## Checklist

### DIFF
--- a/hop/io.py
+++ b/hop/io.py
@@ -88,8 +88,8 @@ class Stream(object):
             in write mode or a :class:`Consumer` instance in read mode.
 
         Raises:
-            ValueError: If the mode is not set to read/write, if more than
-                one topic is specified in write mode, or if more than one broker is specified
+            ValueError: If the mode is not set to read/write, if no topic
+                is specified in read mode, or if more than one broker is specified
 
         """
         username, broker_addresses, topics = kafka.parse_kafka_url(url)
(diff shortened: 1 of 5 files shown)


hop/io.py
```diff
@@ -98,21 +98,20 @@ def open(self, url, mode="r", group_id=None, ignoretest=True, **kwargs):
...
         if self.auth is not None:
             credential = select_matching_auth(self.auth, broker_addresses[0], username)
         else:
             credential = None
 
         if mode == "w":
-            if len(topics) != 1:
-                raise ValueError("must specify exactly one topic in write mode")
+            if topics is None or len(topics) != 1:
+                topics = [None]
```
jnation3406, line 108: Is the idea here that if they set multiple topics, you set the topic to None in the Producer so they are forced to manually enter topic in subsequent write calls so there is no ambiguity?

jnation3406 (approved):
Looks fine to me

hop/io.py
cnweaver: Exactly; the choice we made at the adc level was to store only one topic (https://github.com/astronomy-commons/adc-streaming/blob/b1a5a81c52d2752a9fd63b9717c121285646d09b/adc/producer.py#L89), as either there is a specific topic to which the user wants to always publish or they will have to specify it with ever `write`, and there didn't seem to be much point in artificially restricting the caller in that latter case to one of the predeclared topics, so we just forget what topics they specified except that there was not exactly one.

PR #223, opened 2025-01-14
Implement Large Message Offloading

cnweaver:
## Description

This set of commits adds the integration with the archive API (or any other compatible implementation) to offload messages which are too large to directly transmit through a Kafka topic. 

It also includes an implementation of configuration parameters for users who might want to turn this off, which involves putting in some general machinery for configuration which we didn't really have before.

Known potential issues and things which are not completely implemented:

1. If a (publishing) client checks a topic's message size limit, but the topic then has its size limit decreased, and the client tries to send a message whose size is between the old and new size limits before performing its next periodic settings check, the send will fail instead of being offloaded. Handling this case is not trivial because: 1) detecting the error would involve injecting an error callback, potentially needing to wrap the user's callback, 2) since the error would be discovered asynchronously, resending the message using offloading could alter message order, and 3) because of asynchronous handling of the error, care would be required not to repeat work inefficiently if several messages fail the same way before the first is discovered and corrective action is taken. 
2. Checking the offload server should probably be repeated periodically, so that long-running clients will not fail if the broker changes this. This would have the same problem as with message size limit decreases falling between checks, however, it would be more difficult to detect that a failure is due to stale information rather than some other cause.
3. If a topic has a very low message size limit, it may also be impossible to send the placeholder message after offloading the original. This shouldn't come up much, though, since it is difficult to imagine why anyone would configure a topic with such a low limit (of order a few hundred bytes, or less). 
4. The archive API currently does not return the ID under which a message is stored, so if the client does not label the message with an ID, it will not be able to refer to it. This is mostly not a problem for this implementation, since it generates IDs properly by default, but adds to the complexity any compatible client must have.
5. Neither the archive API nor the message offloading code in this package implements OIDC auth. This would be an issue if a broker which used OIDC wanted to add offloading. 
6. As written, the consumer only understands and accepts BSON payloads returned by the archive API. It might be nice to support fetching arbitrary data, so that users could directly generate messages which get replaced by data hosted at external public URLs? This would pose some potential problems about how data formats should be identified and mapped to model classes, however. 

## Checklist

### DIFF
--- a/hop/configure.py
+++ b/hop/configure.py
@@ -3,6 +3,7 @@ import logging
 import os
 import toml
 from typing import Optional
+import warnings
 
 from . import cli
 
@@ -102,10 +103,16 @@ def load_config(config_path: Optional[str] = None):
         config_path = get_config_path("general")
     if os.path.exists(config_path):
         try:
-            with open(config_path, 'r') as file:
-                config = Config.load(file)
-        except Exception as ex:
-            raise RuntimeError(f"Error loading {config_path}") from ex
+            file = open(config_path, 'r')
+        except Exception:
+            warnings.warn(f"Unable to open {config_path} for reading; using default configuration")
+            config = Config()
+        else:
+            try:
+                with file:
+                    config = Config.load(file)
+            except Exception as ex:
+                raise RuntimeError(f"Error loading {config_path}") from ex
     else:
         config = Config()
 
--- a/hop/configure.py
+++ b/hop/configure.py
@@ -1,5 +1,8 @@
+import dataclasses
 import logging
 import os
+import toml
+from typing import Optional
 
 from . import cli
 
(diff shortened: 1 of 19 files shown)


hop/io.py
```diff
@@ -395,6 +529,129 @@ def read_raw(self, metadata=False, autocommit=True, **kwargs):
...
+            if trusted_offload_url is not None and \
+                    parsed.scheme == trusted_offload_url.scheme and \
+                    parsed.netloc == trusted_offload_url.netloc:
+                auth = http_scram.SCRAMAuth(self.auth, shortcut=True)
+                logger.debug(" Will send auth info in HTTP request")
+        resp = requests.get(url, auth=auth)
+        if not resp.ok:
+            err = _http_error_to_kafka(resp.status_code, f"Failed to fetch message data from {url}:"
+                                       f" HTTP Error {resp.status_code}")
+            if self._conf.error_callback is not None:
```
jnation3406, line 599: I'm not sure where self._conf is set, but could the default for error_callback just be a function that does nothing (i.e. pass)? It would declutter the code a bit since you could remove all the if error_callback is not None stuff.

hop/io.py
```diff
@@ -461,18 +730,160 @@ def __init__(self, broker_addresses, topic, auth, **kwargs):
...
+        logger.debug(f"Fetching settings for topics: {topics}")
+        query = [ConfigResource(restype=ResourceType.TOPIC, name=topic) for topic in topics]
+        futures = aclient.describe_configs(query)
+        results = {}
+        for resource, future in futures.items():
+            try:
+                results[resource.name] = future.result()
+            except confluent_kafka.KafkaException as ke:
+                kerr = ke.args[0]
+                # it is a potentially common issue that a user has WRITE permission but not
```
jnation3406, line 783: Why is this a potentially common issue? I would think it would be good to allow any user to get the max message size?

hop/io.py
```diff
@@ -509,13 +919,106 @@ def write_raw(self, packed_message, headers=None,
...
+            return (None, None, err)
+        if not resp.ok:
+            err = _http_error_to_kafka(resp.status_code,
+                                       "Failed to send large message to offload server at "
+                                       f"{write_url}: POST request failed with status "
+                                       f"{resp.status_code}: {resp.content}")
+            return (None, None, err)
+        msg_url = f"{self.offload_url}/msg/{msg_id}"
+        placeholder = models.ExternalMessage(url=msg_url)
+        # It is important that the placeholder be assigned its own ID, so it does not collide in the
```
jnation3406, line 1015: Would it make sense to have this id be the original message id + some fixed known string, so you can translate between the two easily? I.e. 1234 -> 1234-p for the placeholder?

jnation3406 (approved):
Looks fine to me. Some questions and suggestions, but they may just be due to my lack of understanding all of it.

hop/configure.py
```diff
@@ -37,6 +40,130 @@ def get_config_path(type: str = "general"):
...
+    # First, load the specified file, or the default file if not specified
+    # If the file does not exist, use built-in defaults
+    if config_path is None:
+        config_path = get_config_path("general")
+    if os.path.exists(config_path):
+        try:
+            with open(config_path, 'r') as file:
+                config = Config.load(file)
+        except Exception as ex:
+            raise RuntimeError(f"Error loading {config_path}") from ex
```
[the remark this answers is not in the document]
cnweaver: I had the idea that if a file were specified explicitly it should be an error not to load it, but that's currently rather muddled, as it doesn't necessarily make sense to behave differently for an existing but inaccessible file and a non-existent file, and the logic doesn't distinguish between an explicitly specified file and the default file path, either. I do think it should be treated as a hard error when we open a file and it is malformed, which the current error handling does do. I'll work on figuring out a better way to do this.

hop/io.py
cnweaver: It's set in the constructor, and the default is a function which logs errors (https://github.com/astronomy-commons/adc-streaming/blob/master/adc/consumer.py#L299), but it has been legal for the user to set it to `None`. 

It seems like it would be most obvious to put the `None`->do-nothing-function mapping change in `adc-streaming`, but that would then require a new release of that library. If it seems not too evil, though; I could just do this in the `hop.io.Consumer` constructor, which would avoid needing the adc release dance, and be able to capture a tiny efficiency benefit of letting the `None` be passed down to the lower layers first; I think this gets passed all the way down to the C code, which has a NULL check or equivalent already, and it's faster for it not to call back up into python to make nothing happen.

hop/io.py
cnweaver: I think it would too, but the designers of Kafka made these completely independent permissions, and e.g. our own AuthZ system did not allow querying topic properties by default along with write permission [until I fixed it last week](https://github.com/scimma/kafka-plugins/commit/3b52ee611f816e52716f849e15e43cd0ce9e83a0), so I expect this misconfiguration to exist on many brokers.

hop/io.py
cnweaver: We've currently defined our IDs as UUIDv4, so we don't have a valid way to step out of the 'standard ID namespace' and still have a valid ID. I don't think anything in our design actually depends using UUIDv4, so we could change it, but we would have to plumb that through the archive implementation as well as changing the docs. Also, at the moment it is allowed, (although obviously no one is yet doing it) to have multiple placeholders refer to the same original data. 

So, we could do this, but I'm not sure what we would gain would be worth it.

hop/http_scram.py
```diff
@@ -0,0 +1,164 @@
...
+    def init_per_thread_state(self):
+        if not hasattr(self._thread_local, "init"):
+            self._thread_local.init = True
+            self._thread_local.num_calls = 0
+            self._thread_local.saved_body = None
+
+    def _redo_request(self, r: requests.Response, auth_header: str, final: bool = False, **kwargs):
+        # Consume content and release the original connection
+        # to allow our new request to reuse the same one.
+        r.content
```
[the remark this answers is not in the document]
cnweaver: I'm not sure, but this is the pattern used [in the requests source](https://github.com/psf/requests/blob/main/src/requests/auth.py#L265-L268), so it seemed best to follow it. I at least hope they had a good reason, since I can't get any visible failure to happen by taking this out, but I'm not confident that means it wouldn't fail in some other circumsatnces I haven't tested.

hop/http_scram.py
```diff
@@ -0,0 +1,164 @@
...
+from scramp import ScramClient
+import secrets
+import threading
+
+
+logger = logging.getLogger("hop")
+
+
+class SCRAMAuth(requests.auth.AuthBase):
+    def __init__(self, credential, shortcut: bool = False, check_final=True):
```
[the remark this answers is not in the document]
cnweaver: Thanks for the catch; this should have been documented, which I have now added.

hop/io.py
jnation3406: Yeah that sounds okay to me

hop/configure.py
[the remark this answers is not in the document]
cnweaver: This is now implemented.

PR #225, opened 2025-02-18
Write message models to stdout in their original formats.

cnweaver:
The goal is that our code should be able to write models back out in their expected formats so that users can handle them normally with other tools, and so that our own model code is capable of reading them back in. 
This removes random python repr cruft from blob messages, shows GCN Text Notices in their original form, and outputs AvroBlobs as valid Avro. VOEvent is not yet handled because the importation from XML is lossy with regard to which bits are elements or attributes, and these need to be restored correctly according to the schema. 

For most models this also leads to simpler, clearer code, as `__bytes__` is responsible for serializing the data to a a flat form, and `serialize` is responsible for packaging that form together with the format name for the Kafka machinery, so most models do not need their own implementations of it.

No attempt is made in this patch to alter the fact that some models use bespoke formats for transmission over Kafka. 

## Checklist

### DIFF
--- a/hop/models.py
+++ b/hop/models.py
@@ -28,8 +28,13 @@ class MessageModel(ABC):
 
     """
 
+    def __bytes__(self):
+        """Produce the canonical representation of this message.
+        """
+        return json.dumps(asdict(self)).encode("utf-8")
+
     def serialize(self):
-        """Wrap the message with its format and content.
+        """Wrap the message with its format and content, for transmission of Kafka.
 
         Returns:
             A dictionary with "format" and "content" keys.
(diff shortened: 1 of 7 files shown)


adambrazier (approved):
This looks good; I tested sending all of the files in tests/data/test_data with both the correct filestype with the `-f` option for each to publish and also the default BLOB. I wrote to `stdout` and also piped to file so I could diff

All the tests sent as BLOB came out identical with the source file

Using `-f`, voevent doesn't come out right, which is a known issue we will fix in a future branch. the blob text and both the gcn types come through identical as measured with diff. JSON has the same content but shows a non-null diff; this will produce the same python objects and we think is due to whitespace choices by the formatter. AVRO is binary and does show a non-null diff but Chris tested and it also produces the same python objects

